PR #15, opened 2018-02-06
Add support for node modules, Add node project template, and node mod…

HadwaAbdelhalem:
…ules test. Updated filter module template to include module.json


iotedgedev/modules.py
```diff
@@ -27,16 +29,20 @@ def build(self):
 
                 self.output.info("BUILDING MODULE: {0}".format(module_dir))
 
-                # Find first proj file in module dir and use it.
-                project_files = [os.path.join(module_dir, f) for f in os.listdir(
-                    module_dir) if f.endswith("proj")]
+                # if no module.json found ? default to .net
+                json_file_path = os.path.join(os.getcwd(), "modules", module,"module.json")
```
jonbgallant: module_dir has the path to module, you need to do os.path.join(module_dir, "module.json")

jonbgallant: Please refactor this to create a Module class that can init via JSON object or a file path to a JSON file.  It will expose the JSON object from the modules.json file and have defaults in it.

That way later in code you can do things like module.language and so on...

iotedgedev/modules.py
```diff
@@ -27,16 +29,20 @@ def build(self):
...
+                    file_json_content = json.loads (self.utility.get_file_contents(json_file_path))
+                    module_lang = file_json_content.get("language").lower()
 
-                if len(project_files) == 0:
-                    self.output.error("No project file found for module.")
-                    continue
+                else :
+                    module_lang = "csharp"
+                
+                mod_proc_facroty = ModulesProcessorFactory(self.envvars, self.utility, self.output, self.dock)
```
jonbgallant: spelling...factory

iotedgedev/modules.py
jonbgallant: Please make sure you run ALT+SHIFT+F to format document.  You'll need to install the VS Code extension and autopep8.  Install extension, run Format once and it will prompt you to install autopep8

iotedgedev/modules.py
```diff
@@ -163,3 +168,31 @@ def deploy_device_configuration(
             self.output.error(
                 "There was an error deploying the configuration. Please make sure your IOTHUB_CONNECTION_STRING and DEVICE_CONNECTION_STRING Environment Variables are correct.")
 
+class ModulesProcessorFactory(Modules):
+    
+    def factory(self, module_language):
```
jonbgallant: If you are going to use Factory in the class name, then rename the method to "get"

iotedgedev/modules.py
```diff
@@ -163,3 +168,31 @@ def deploy_device_configuration(
...
+        if module_language == "csharp" or  module_language == "fsharp" or module_language == "vbasic":
```
jonbgallant: If you do a compare like this you want to make sure they are the same casing.  I know you called to Lower before, but this class shouldn't assume that is done.

iotedgedev/modules.py
```diff
@@ -163,3 +168,31 @@ def deploy_device_configuration(
...
+            return DotNetModuleProcessor(self.envvars, self.utility, self.output, self.dock)
+
+        else :
+            return OtherModuleProcessor(self.envvars, self.utility, self.output, self.dock)
+
+class DotNetModuleProcessor(ModulesProcessorFactory):
+    def build(self, module_dir):
+        project_file = [os.path.join(module_dir, f) for f in os.listdir(
```
jonbgallant: What happens if the project file doesn't exist?

iotedgedev/modules.py
```diff
@@ -163,3 +168,31 @@ def deploy_device_configuration(
...
+                    module_dir) if f.endswith("proj")][0]
+        self.utility.exe_proc(["dotnet", "build", project_file,
+                                       "-v", self.envvars.DOTNET_VERBOSITY])
+
+    def publish(self, module_dir, build_path):
+        project_file = [os.path.join(module_dir, f) for f in os.listdir(
```
jonbgallant: Same, don't assume project file exists

iotedgedev/template/modules/filter-module/module.json
```diff
@@ -0,0 +1,12 @@
+{
```
jonbgallant: Please update to use this schema: https://github.com/Azure/dotnet-template-azure-iot-edge-module/blob/master/content/dotnet-template-azure-iot-edge-module/CSharp/module.json

tests/node-project/modules/node-module/module.json
```diff
@@ -0,0 +1,12 @@
+{
```
jonbgallant: Update: https://github.com/Azure/dotnet-template-azure-iot-edge-module/blob/master/content/dotnet-template-azure-iot-edge-module/CSharp/module.json

HadwaAbdelhalem: the link is not working. I see 404 page

iotedgedev/modules.py
```diff
@@ -163,3 +160,58 @@ def deploy_device_configuration(
...
+        self.envvars = envvars
+        self.utility = utility
+        self.output = output
+        self.module_json_file = module_json_file
+
+    def load_module_json(self):
+        if os.path.exists(self.module_json_file):
+            file_json_content = json.loads(
+                self.utility.get_file_contents(self.module_json_file))
+            return file_json_content.get("language")
```
jonbgallant: This should be the whole json object, not just language.

iotedgedev/modules.py
```diff
@@ -163,3 +160,58 @@ def deploy_device_configuration(
...
+
+        else:
+            self.output.info(
+                "No module.json file found. Default to dotnet module")
+            return "csharp"
+
+    def get(self):
+        module_language = self.load_module_json().lower()
```
jonbgallant: Then here you get language

PR #16, opened 2018-02-09
Move docker files to root

HadwaAbdelhalem:


iotedgedev/module.py
```diff
@@ -0,0 +1,40 @@
...
+    def __init__(self, output, utility, module_json_file):
+        self.output = output
+        self.utility = utility
+        self.module_json_file = module_json_file
+
+        self.module_language = "csharp"
+        self.file_json_content = None
+        self.load_module_json()
+
+    def load_module_json(self):
```
jonbgallant: Actually, if module.json not found or there's an error, display a message that says 
not found: "module.json file is required in the root of your module folder."
error: "error while loading module.json file: {path}"  (use self.output.error())

jonbgallant: With this update we are not requiring module.json file...

iotedgedev/modules.py
```diff
@@ -70,44 +83,39 @@ def build(self):
...
 
-                        tag_name = runtime + ext + container_tag
+                        #tag_name = runtime + ext + container_tag
+                        tag_name = module_json.tag_version + container_tag
 
                         # construct the build output path
                         build_path = os.path.join(
-                            os.getcwd(), "build", "modules", module, runtime)
+                            os.getcwd(), "build", "modules", module)  # , runtime)
                         if not os.path.exists(build_path):
```
jonbgallant: This isn't needed.

iotedgedev/modules.py
```diff
@@ -70,44 +83,39 @@ def build(self):
...
                         build_dockerfile = os.path.join(
                             build_path, docker_file_name)
-
-                        copyfile(docker_file, build_dockerfile)
+                        copyfile(os.path.join(
+                            module_dir, docker_file_name), build_dockerfile)
 
                         image_destination_name = "{0}/{1}:{2}".format(
-                            self.envvars.CONTAINER_REGISTRY_SERVER, module, tag_name).lower()
+                            self.envvars.CONTAINER_REGISTRY_SERVER, module, module_json.tag_version, "-", arch_key, tag_name).lower()
```
jonbgallant: This doesn't seem right.  It should be modulename:version-containertag-arch.

Also with format...you only have 3 placeholders, but 6 parameters.

iotedgedev/modules.py
```diff
@@ -121,7 +129,7 @@ def build(self):
                         self.output.info(
                             "PUSHING DOCKER IMAGE TO: " + image_destination_name)
 
-                        for line in self.dock.docker_client.images.push(repository=image_destination_name, tag=tag_name, stream=True, auth_config={
+                        for line in self.dock.docker_client.images.push(repository=image_destination_name, stream=True, auth_config={
```
jonbgallant: Did you find that tag_name is not required?

iotedgedev/template/deployment.template.json
```diff
@@ -45,7 +45,7 @@
             "status": "running",
             "restartPolicy": "always",
             "settings": {
-              "image": "${CONTAINER_REGISTRY_SERVER}/filter-module:linux-x64-${CONTAINER_TAG}",
```
jonbgallant: Let's remove container tag from the default template.  We'll document how to use it.

tests/node-project/deployment.template.json
```diff
@@ -0,0 +1,74 @@
...
+              "createOptions": ""
+            }
+          },
+          "node-module": {
+            "version": "1.0",
+            "type": "docker",
+            "status": "running",
+            "restartPolicy": "always",
+            "settings": {
+              "image": "${CONTAINER_REGISTRY_SERVER}/node-module:linux-x64-${CONTAINER_TAG}",
```
jonbgallant: Remove container tag here,

tests/node-project/deployment.template.json
```diff
@@ -0,0 +1,74 @@
...
+            "version": "1.0",
+            "type": "docker",
+            "status": "running",
+            "restartPolicy": "always",
+            "settings": {
+              "image": "microsoft/azureiotedge-simulated-temperature-sensor:${RUNTIME_TAG}",
+              "createOptions": ""
+            }
+          },
+          "node-module": {
```
jonbgallant: change name to nodemodule .... in every file/reference of it across thh board

tests/test_iotedgedev.py
```diff
@@ -109,4 +113,47 @@ def test_no_dotenv_file(self):
...
+        result = runner.invoke(cli.main, ['modules', '--build'])
+        print(result.output)
+        assert result.exit_code == 0  
+
+        result = runner.invoke(cli.main, ['modules', '--deploy'])
+        print(result.output)
+        assert result.exit_code == 0
+        
+        src = os.path.join(root_dir, node_project, "modules")
+        dist = os.path.join(root_dir, node_project, "build", "modules")
```
jonbgallant: Build folder will now be gone. So we need to think of another way to test this.

iotedgedev/dotnetmoduleprocessor.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class DotNetModuleProcessor(object):
+    def __init__(self, envvars, utility, output):
+        self.envvars = envvars
+        self.utility = utility
+        self.output = output
+
+    def build(self, module_dir):
+        project_files = [os.path.join(module_dir, f) for f in os.listdir(
```
jonbgallant: This code is used in both build and publish.  Please refactor into a method that returns the first project file given module_dir.

iotedgedev/module.py
```diff
@@ -0,0 +1,49 @@
...
+
+                self.module_language = self.file_json_content.get(
+                    "language").lower()
+            except:
+                self.output.error(
+                    "Error while loading module.json file : {0}".format(self.module_json_file))
+
+        else:
+            self.output.error(
+                "No module.json file found. module.json file is required in the root of your module folder")
```
jonbgallant: I think a sys.exit() is going to be required here to stop the script from continuing to execute.

iotedgedev/module.py
```diff
@@ -0,0 +1,49 @@
...
+        self.utility = utility
+        self.module_json_file = module_json_file
+
+        self.module_language = "csharp"
+        self.file_json_content = None
+        self.load_module_json()
+
+    def load_module_json(self):
+        if os.path.exists(self.module_json_file):
+            try:
```
jonbgallant: We should do some more schema validation here, I'll add a task for a future update.

iotedgedev/modules.py
```diff
@@ -27,86 +31,48 @@ def build(self):
 
                 self.output.info("BUILDING MODULE: {0}".format(module_dir))
 
-                # Find first proj file in module dir and use it.
-                project_files = [os.path.join(module_dir, f) for f in os.listdir(
-                    module_dir) if f.endswith("proj")]
+                module_json = Module(
```
jonbgallant: call this module instead of module_json

iotedgedev/module.py
```diff
@@ -0,0 +1,49 @@
+
+import os
+import json
+
+
+class Module(object):
+    def __init__(self, output, utility, module_json_file):
```
jonbgallant: I would prefer to pass module_dir into the ctor, then you don't have to pass to each method.

iotedgedev/modules.py
```diff
@@ -27,86 +31,48 @@ def build(self):
...
-                        build_dockerfile = os.path.join(
-                            build_path, docker_file_name)
-
-                        copyfile(docker_file, build_dockerfile)
-
-                        image_destination_name = "{0}/{1}:{2}".format(
-                            self.envvars.CONTAINER_REGISTRY_SERVER, module, tag_name).lower()
+                        mod_proc.publish(module_dir)
+                        image_destination_name = "{0}/{1}:{2}-{3}".format(
+                            self.envvars.CONTAINER_REGISTRY_SERVER, module, tag_name, docker_file_name).lower()
```
jonbgallant: Won't this be filtermodule:0.0.0-Dockerfile?  I think you mean "arch" here and not docker_file_name

iotedgedev/template/deployment.template.json
```diff
@@ -39,13 +39,13 @@
...
           },
-          "filter-module": {
+          "filtermodule": {
             "version": "1.0",
             "type": "docker",
             "status": "running",
             "restartPolicy": "always",
             "settings": {
-              "image": "${CONTAINER_REGISTRY_SERVER}/filter-module:linux-x64-${CONTAINER_TAG}",
+              "image": "${CONTAINER_REGISTRY_SERVER}/filtermodule:amd64",
```
jonbgallant: This should have a version number

tests/nodeproject/deployment.template.json
```diff
@@ -0,0 +1,74 @@
...
+              "createOptions": ""
+            }
+          },
+          "node-module": {
+            "version": "1.0",
+            "type": "docker",
+            "status": "running",
+            "restartPolicy": "always",
+            "settings": {
+              "image": "${CONTAINER_REGISTRY_SERVER}/nodemodule:amd64${CONTAINER_TAG}",
```
jonbgallant: Needs version number

iotedgedev/modules.py
HadwaAbdelhalem: that would conflict with the module var in for loop

iotedgedev/modules.py
HadwaAbdelhalem: I do not believe it is

iotedgedev/module.py
jonbgallant: I meant to say we **are** requiring the module.json file.

iotedgedev/modules.py
jonbgallant: Why did you remove it?

iotedgedev/module.py
HadwaAbdelhalem: yes , and I added sys.exit() after logging error message if the file was missing.(Module.py line #32)

Issue #20, opened 2018-02-11
With addition of az cli, the tool takes a long time to install


jonbgallant:
Can we trim the az-cli to the required modules for this scenario?

PR #29, opened 2018-02-27
Lightweight Az CLI via pip install method.

digimaun:
Addresses issue #20.


paloukari:
This is what I initially tried when referencing az cli, but it's not enough. Try uninstalling az cli from your environment and doing an e2e test. The token cache seems not to function as expected.

My guess is that either there is a missing setup step, or a missing package.

PR #158, opened 2018-04-18
Added support for windows containers via nanoserver

yorek:


docker/linux/Dockerfile
```diff
@@ -24,4 +24,4 @@ RUN pip --no-cache-dir install -U iotedgedev
 WORKDIR /home/iotedge
 COPY install-dev.sh /scripts/install-dev.sh
 RUN sed -i 's/\r//' /scripts/install-dev.sh
-ENV DEBIAN_FRONTEND teletype
\ No newline at end of file
+ENV DEBIAN_FRONTEND teletype
```
jonbgallant: What is the change here?  If no changes, then please remove from the commit.  Thanks,

docker/windows/Dockerfile
```diff
@@ -0,0 +1,48 @@
+FROM microsoft/windowsservercore:1709 AS base
+SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
+ENV PYTHON_VERSION 3.6.5
+ENV NODEJS_VERSION 8.11.1
+ENV DOTNETCORESDK_VERSION 2.1.4
```
jonbgallant: can you also upgrade the versions in the linux Dockerfile to match this Windows Dockerfile?

docker/windows/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+    
+FROM microsoft/nanoserver:1709
+RUN mkdir c:\\tools 
+COPY --from=base ["tools/", "/tools"]
+USER ContainerAdministrator
+RUN setx /M PATH "%PATH%;C:\tools\dotnetcoresdk;C:\tools\node\;C:\tools\python;C:\tools\python\Scripts;c:\tools\docker\;"
+USER ContainerUser
+RUN python -m pip install --upgrade pip
+RUN pip install azure-cli
+RUN npm i -g iothub-explorer
```
jonbgallant: I also had to `npm i -g tree-kill`

docker/windows/install-dev.bat
```diff
@@ -0,0 +1,4 @@
+cd /home/iotedge/tool
```
jonbgallant: Can we create an install-dev file that works cross plat?  So we don't need two of them?

docker/windows/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+export VERSION=$1
+docker build -f Dockerfile -t microsoft/iotedgedev:$1 -t microsoft/iotedgedev:$1-nanoserver -t microsoft/iotedgedev:latest-nanoserver .
```
jonbgallant: Let's tag with windows instead of nanoserver.  I doubt we'll ever have another version of windows container.

docker/windows/run.ps1
```diff
@@ -0,0 +1,9 @@
+if ((Get-NetFirewallRule -DisplayName "Docker Proxy" -ErrorAction Ignore) -eq $null)
```
jonbgallant: Please add a comment to this file that they need to enable the expose daemon setting in Docker.

docker/windows/run.ps1
```diff
@@ -0,0 +1,9 @@
...
+{
+    New-NetFirewallRule -DisplayName "Docker Proxy" -LocalPort 2375 -Action Allow -Protocol TCP
+}
+
+$docker_gateway = (Get-NetIPAddress -InterfaceAlias 'vEthernet (nat) 2' -AddressFamily IPv4 | Select-Object IPAddress).IPAddress
+$docker_host ="tcp://{0}:2375" -f $docker_gateway
+
+docker run -it -e DOCKER_HOST=$docker_host microsoft/iotedgedev:latest-nanoserver
```
jonbgallant: update to windows after you fix the tag.

jonbgallant: add the same volume mount c:/temp/iotedge:/home/iotedge

docker/windows/run_dev.ps1
```diff
@@ -0,0 +1,12 @@
+if ((Get-NetFirewallRule -DisplayName "Docker Proxy" -ErrorAction Ignore) -eq $null)
```
jonbgallant: I don't think this file is needed.

docker/windows/Dockerfile
```diff
@@ -0,0 +1,48 @@
+FROM microsoft/windowsservercore:1709 AS base
+SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
+ENV PYTHON_VERSION 3.6.5
+ENV NODEJS_VERSION 8.11.1
+ENV DOTNETCORESDK_VERSION 2.1.4
+ENV DOCKER_VERSION 17.09.0
```
jonbgallant: Can you also add Python 2.7?

docker/linux/Dockerfile
yorek: Yep

docker/windows/Dockerfile
yorek: I think we should put all version in an external file for easier maintenance

docker/windows/Dockerfile
yorek: Added

docker/windows/install-dev.bat
yorek: Unfortunately not since nanoserver doesn't support .sh scripts

docker/windows/build.sh
yorek: Done

docker/windows/run.ps1
yorek: Ok

docker/windows/run.ps1
yorek: Done

yorek: Done

docker/windows/run_dev.ps1
yorek: Removed

docker/windows/Dockerfile
yorek: Yes, but it will be complex since Python 2.7 and Python 3.6 use two different installers :(. I'd like to be able two support the two versions without having to maintain also two dockerfiles

yorek: Yep, make sense

.gitignore
```diff
@@ -84,4 +84,5 @@ node_modules
 
 /docker/linux/Dockerfile.expanded
```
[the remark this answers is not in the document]
yorek: Done

jonbgallant:
I get this when I run ./build.sh ... should we upgrade pip in the container first?

You are using pip version 8.1.1, however version 10.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.

setup.py
```diff
@@ -53,7 +53,7 @@ def run(self):
 
 setup(
     name='iotedgedev',
-    version='0.81.0',
+    version='0.86.0',
```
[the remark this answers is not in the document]
yorek: Yep, it slipped through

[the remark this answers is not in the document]
yorek: The build script will execute bumpversion only when building for production and will not execute it it building a test release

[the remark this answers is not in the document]
yorek: script done

yorek:
Finally had python 3.6 as default python3 version working, but had to switch to ubuntu 18.04. 
When tox is run from within the Linux container (haven't tried yet from the windows continers) it now breaks with a .NET Error:

```
BUILDING MODULE: modules/filtermodule

ERROR: System.ArgumentNullException: Value cannot be null.
Parameter name: path1
   at System.IO.Path.Combine(String path1, String path2)
   at Microsoft.DotNet.Configurer.CliFallbackFolderPathCalculator.get_DotnetUserProfileFolderPath()
   at Microsoft.DotNet.Configurer.FirstTimeUseNoticeSentinel..ctor(CliFallbackFolderPathCalculator cliFallbackFolderPathCalculat
or)
   at Microsoft.DotNet.Cli.Program.ProcessArgs(String[] args, ITelemetry telemetryClient)
   at Microsoft.DotNet.Cli.Program.Main(String[] args)
None

```
The same error doesn't happen if tox is executed in the host, outside the container. .NET Core SDK in host and container both to 2.1.200

jongio:
@yorek 
As discussed:
1. Break Dockerfile into two parts so we can use cache of all the dependencies. Dockerfile.base..for each OS
2. Use the local wheel instead of pypi when building container.

yorek:
All done :)

yorek:
All done :)

PR #162, opened 2018-04-19
check if dir is empty on init and prompt accordingly

### LINKED ISSUE #155: init - make sure folder is empy
prompt if not empty and ask user to run azure setup or clear the directory

yaron2:
Fixes #155

### DIFF
--- a/iotedgedev/cli.py
+++ b/iotedgedev/cli.py
@@ -74,13 +74,16 @@ def solution(create, name):
 @click.pass_context
 def init(ctx):
 
+    if len(os.listdir(os.getcwd())) > 0:
+        output.prompt("Directory is not empty. Run iotedgedev azure --setup or clear the directory.")
+        sys.exit()
+
     utility = Utility(envvars, output)
 
-    if len(os.listdir(os.getcwd())) == 0:
-        solcmd = "iotedgedev solution ."
-        output.header(solcmd)
-        utility.call_proc(solcmd.split())
-
+    solcmd = "iotedgedev solution ."
+    output.header(solcmd)
+    utility.call_proc(solcmd.split())
+    
     azsetupcmd = "iotedgedev azure --update-dotenv"
     output.header(azsetupcmd)
     utility.call_proc(azsetupcmd.split())
--- a/iotedgedev/cli.py
+++ b/iotedgedev/cli.py
@@ -74,16 +74,13 @@ def solution(create, name):
 @click.pass_context
 def init(ctx):
 
-    if len(os.listdir(os.getcwd())) > 0:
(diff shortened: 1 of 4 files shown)


iotedgedev/cli.py
```diff
@@ -74,13 +74,16 @@ def solution(create, name):
 @click.pass_context
 def init(ctx):
 
-    utility = Utility(envvars, output)
```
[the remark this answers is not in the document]
yaron2: ok cool. thought you wanted to limit the scope of the check to init because of the #155 title :)

iotedgedev/solution.py
```diff
@@ -26,3 +30,14 @@ def create(self, name):
         self.output.footer("Azure IoT Edge Solution Created")
         if name != "":
             self.output.info("Execute 'cd {0}' to navigate to your new solution.".format(name))
+
+    def is_dir_empty(name):
```
jonbgallant: Please put this in Utility.py as we may need it in other places later.

Can you also create tests for these use cases?

tests/test_iotedgedev.py
```diff
@@ -41,6 +41,33 @@ def clean():
     return
 
 
+@pytest.fixture
+def test_solution_create_in_non_empty_dir(request):
```
jonbgallant: Please also test the current folder path "iotedgedev solution ."

iotedgedev/solution.py
```diff
@@ -1,12 +1,16 @@
...
-
 class Solution:
-    def __init__(self, output):
+    def __init__(self, output, utility):
         self.output = output
+        self.utility = utility
 
     def create(self, name):
+        if not self.utility.is_dir_empty(name):
+            self.output.prompt("Directory is not empty. Run 'iotedgedev azure' or clean the directory")
```
jonbgallant: Add "." at end of the sentence.

iotedgedev/utility.py
```diff
@@ -41,6 +41,17 @@ def call_proc(self, params, shell=False):
         except Exception as e:
             self.output.error("Error while executing command: {0}. {1}".format(' '.join(params), str(e)))
 
+    def is_dir_empty(self, name):
+        if name == ".":
```
jonbgallant: I feel like lines 45-48 where you resolve name to full name shouldn't be in this method called is_dir_empty.  I think that logic can be in the create method for now.

yaron2: Done. I'd actually keep that logic in is_empty_dir because although its a utility method I think it can well encapsulate the actual name-to-path manipulation, but that's just an implementation detail really.

PR #169, opened 2018-05-03
Added vscode to .gitignore

yaron2:
Globally shared vscode settings can mess up a dev environment, especially the current python.pythonPath is not friendly towards us Mac users...


jonbgallant:
I have that there for styling reasons, so users format files the same.  Please clean up anything in the file versus removing it.  Thanks.

jongio:
@yaron2 - Can you update to resolve conflicts?

yaron2:
@jongio Done

PR #173, opened 2018-05-16
AZ IOT HUB apply-configuration needs hubname

### LINKED ISSUE #167: AZ IOT HUB apply-configuration needs hubname.
If user has old version of az cli iot extension installed they get this: 

`az iot hub apply-configuration: error: argument --hub-name/-n is required
`

- add the -n parameter to the apply-configuration call.  you can get it in IOTHUB_CONNECTION_INFO.HostName

apply-configuration might need ONLY hubname, but HostName has [name].azuredevices.net.

Therefore, You might have to split the ConnectionString.HostName property and add a new property to that class called HubName.

kln4dt:
this commit fixes #167

### DIFF
--- a/iotedgedev/azurecli.py
+++ b/iotedgedev/azurecli.py
@@ -226,10 +226,10 @@ class AzureCli:
 
         return result
 
-    def apply_configuration(self, deviceId, connection_string, hubName, config):
-        self.output.status(f("Deploying '{config}' to '{deviceId}'..."))
+    def apply_configuration(self, device_id, connection_string, hub_name, config):
+        self.output.status(f("Deploying '{config}' to '{device_id}'..."))
 
-        return self.invoke_az_cli_outproc(["iot", "hub", "apply-configuration", "-d", deviceId, "-n", hubName, "-k", config, "-l", connection_string], error_message=f("Failed to deploy '{config}' to '{deviceId}'..."), suppress_output=True)
+        return self.invoke_az_cli_outproc(["iot", "hub", "apply-configuration", "-d", device_id, "-n", hub_name, "-k", config, "-l", connection_string], error_message=f("Failed to deploy '{config}' to '{device_id}'..."), suppress_output=True)
 
     def get_free_iothub(self):
         with output_io_cls() as io:
--- a/iotedgedev/azurecli.py
+++ b/iotedgedev/azurecli.py
@@ -226,10 +226,10 @@ class AzureCli:
 
         return result
 
-    def apply_configuration(self, deviceId, connection_string, config):
+    def apply_configuration(self, deviceId, connection_string, hubName, config):
         self.output.status(f("Deploying '{config}' to '{deviceId}'..."))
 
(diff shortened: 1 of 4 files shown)


iotedgedev/azurecli.py
```diff
@@ -226,10 +226,10 @@ def list_resource_groups(self):
 
         return result
 
-    def apply_configuration(self, deviceId, connection_string, config):
+    def apply_configuration(self, deviceId, connection_string, hubName, config):
```
jonbgallant: I just noticed that the parameter names here are a mix of camelCase and underscores.  Can you make consistent with underscores?

`(self, device_id, connection_string, hub_name, config)`

iotedgedev/edge.py
```diff
@@ -14,7 +14,7 @@ def deploy(self):
         self.envvars.verify_envvar_has_val("DEVICE_CONNECTION_STRING", self.envvars.DEVICE_CONNECTION_STRING)
         self.envvars.verify_envvar_has_val("DEPLOYMENT_CONFIG_FILE", self.envvars.DEPLOYMENT_CONFIG_FILE)
 
-        self.azure_cli.apply_configuration(self.envvars.DEVICE_CONNECTION_INFO.DeviceId, self.envvars.IOTHUB_CONNECTION_STRING, self.envvars.DEPLOYMENT_CONFIG_FILE_PATH)
+        self.azure_cli.apply_configuration(self.envvars.DEVICE_CONNECTION_INFO.DeviceId, self.envvars.IOTHUB_CONNECTION_STRING, self.envvars.IOTHUB_CONNECTION_INFO.HubName, self.envvars.DEPLOYMENT_CONFIG_FILE_PATH)
```
jonbgallant: 1. Change the check above to IOTHUB_CONNECTION_INFO
`self.envvars.verify_envvar_has_val("IOTHUB_CONNECTION_INFO", self.envvars.IOTHUB_CONNECTION_INFO)`

2. Change following to use IOTHUB_CONNECTION_INFO

`self.azure_cli.apply_configuration(self.envvars.DEVICE_CONNECTION_INFO.DeviceId, self.envvars.IOTHUB_CONNECTION_INFO.ConnectionString, self.envvars.IOTHUB_CONNECTION_INFO.HubName, self.envvars.DEPLOYMENT_CONFIG_FILE_PATH)`

iotedgedev/connectionstring.py
```diff
@@ -13,6 +13,8 @@ def __init__(self, value):
 
             if self.data:
                 self.HostName = self["hostname"]
+                if self.HostName:
```
jonbgallant: Keep `def __init__(self, value)` as is.

Change self.value to self.ConnectionString

`self.ConnectionString = value`

And then update all references to self.value to self.ConnectionString.

Also, line 7 above references value.split(';'), that should be self.ConnectionString.split(';')

iotedgedev/edge.py
kln4dt: shouldn't the check for device connection string be changed to DEVICE_CONNECTION_INFO too?

jonbgallant: Yes.

PR #174, opened 2018-05-18
Lowercase the search token for subscription ids, and handle ambiguous or empty results

druttka:
- When searching for the subscription id, lowercase the search token
- When multiple subscriptions match, do not assume the first was desired

### DIFF
--- a/iotedgedev/azurecli.py
+++ b/iotedgedev/azurecli.py
@@ -159,11 +159,19 @@ class AzureCli:
         with output_io_cls() as io:
             result = self.invoke_az_cli_outproc(["account", "list", "--query", "[?starts_with(@.id,'{0}') || starts_with(@.name,'{1}')] | [0]".format(token.lower(), token)],
                                                 "Could not find a subscription that starts with '{0}'".format(token), io)
+
             if result:
                 out_string = io.getvalue()
                 if out_string:
+
                     data = json.loads(out_string)
-                    return data["id"]
+                    if len(data) == 1:
+                        return data[0]["id"]
+                    elif len(data) > 1:
+                        self.output.error("Found multiple subscriptions that start with '{0}'".format(token))
+                    else:
+                        self.output.error("Could not find a subscription that starts with '{0}'".format(token))
+
         return ''
 
     def set_subscription(self, subscription):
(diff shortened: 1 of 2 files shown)


iotedgedev/azurecli.py
```diff
@@ -157,24 +157,35 @@ def get_default_subscription(self):
...
+                    elif len(data) > 1:
+                        self.output.error("Found multiple subscriptions that start with '{0}'".format(token))
+                    else:
+                        self.output.error("Could not find a subscription that starts with '{0}'".format(token))
+
         return ''
 
     def set_subscription(self, subscription):
 
         if len(subscription) < 36:
```
druttka: I considered adding a new issue to handle more use cases / user experience. 

For examples, 

- what if they were searching by name and happened to provide >= 36 char
- should we support contains instead of starts with
- can we make this even more user friendly, e.g., provide integer ids for each subscription and prompt to select 1, 2, 3...
- 

I decided any of those could be just as well done as separate issues after decided what we want.

iotedgedev/azurecli.py
```diff
@@ -157,24 +157,35 @@ def get_default_subscription(self):
...
             if result:
                 out_string = io.getvalue()
                 if out_string:
+
                     data = json.loads(out_string)
-                    return data["id"]
+                    if len(data) == 1:
+                        return data[0]["id"]
+                    elif len(data) > 1:
+                        self.output.error("Found multiple subscriptions that start with '{0}'".format(token))
```
jongio: Found multiple subscriptions that start with x. Please enter more characters to further refine your selection.

iotedgedev/azurecli.py
jongio: 1. yes, checking bounds
2. no, want starts with because less likely chance for multiples returned
3. yes, but no easy way to do it.

Agree that these should be sep issues.

druttka: Just a thought for 2. When searching by name, if I have "SE Account - david" and "SE Account - jon", then contains "dav" is less likely to return multiples than starts with "SE Account - ".

jongio:
Can you add tests for this?

druttka:
Can you point me to an existing harness or example of mocking the az account output when invoking get_subscription_id_starts_with? Otherwise I'll need some time to investigate how to make get_subscription_id_starts_with testable.

druttka:
@jongio What do you think about the added test_azurecli. It doesn't do a full mocking/faking of az cli output for testing the zero, one, or multiple results flow yet, but does assert how the search token becomes a JMESPath query. Manual testing different combos of zero, one, multiple results with lower, upper, mixed case worked for me too.

jongio:
@druttka I like it.  Also, I think we are good to change this to contains if you are ready to do so.

druttka:
Thanks. I won't get to contains today (just switched contexts to verify the fail fast dotnet prompt). So I can handle that separately as opened in #177 if you want to get this part in, or can combine #177 into this if you want to wait.

jongio:
I'm okay waiting.   Will probably save us both some time to do at once.

druttka:
@jongio  code and tests updated to use contains for the name search. Output messages also changed as below

### When there are no matches
ERROR: Could not find a subscription for which the id starts with or name contains 'foo'.

### When there are multiple matches
ERROR: Found multiple subscriptions for which the ids start with or names contain 'O'. Please enter more characters to further refine your selection.

Issue #95, opened 2018-03-20
add create module command


jonbgallant:

jonbgallant:
API: 
`iotedgedev module [modulename] --lang [lang, default=c#]

PR #180, opened 2018-05-28
Add commands to add module to solution

LazarusX:
Work-in-progress for #95 

Remaining tasks:
* Validate input
* Add tests
* Add CLI entry point
* Handle errors
* Update README
* etc.

------
Updated in June 13:
Most of above tasks are implemented in this PR


iotedgedev/cli.py
```diff
@@ -423,6 +423,25 @@ def azure(setup,
             envvars.save_envvar("DEVICE_CONNECTION_STRING", envvars.DEVICE_CONNECTION_STRING)
             output.info("Updated current .env file")
 
+# @click.command(context_settings=CONTEXT_SETTINGS, help="Create IoT Edge Modules")
```
LazarusX: This is just for temporary testing purpose. I will revisit the CLI entry point design later.

iotedgedev/modules.py
```diff
@@ -15,6 +14,51 @@ def __init__(self, envvars, utility, output, dock):
         self.dock = dock
         self.dock.init_registry()
 
+    def create(self, name, lang):
```
adashen: how about change input name "lang" to "type"? the value specify on module type not language. (Consider csharp custom module and csharp function module)

iotedgedev/modules.py
```diff
@@ -15,6 +14,51 @@ def __init__(self, envvars, utility, output, dock):
...
+        self.output.header("CREATING MODULE")
+
+        if lang == "csharp":
+            cmd = "dotnet new -i Microsoft.Azure.IoT.Edge.Module"
+            self.output.header(cmd)
+            self.utility.exe_proc(cmd.split())
+            cmd = "dotnet new aziotedgemodule -n {0} -r {1}".format(name, self.envvars.CONTAINER_REGISTRY_SERVER)
```
adashen: use CONTAINER_REGISTRY_SERVER as repository name?

iotedgedev/modules.py
```diff
@@ -15,6 +14,51 @@ def __init__(self, envvars, utility, output, dock):
...
+            self.output.header(cmd)
+            self.utility.exe_proc(cmd.split(), cwd=self.envvars.MODULES_PATH)
+        elif lang == "nodejs":
+            pass
+        elif lang == "python":
+            github_source = "https://github.com/Azure/cookiecutter-azure-iot-edge-module"
+            branch = "master"
+            cmd = ("cookiecutter --no-input {0} module_name={1} image_repository={2} --checkout {3}"
+                   .format(github_source, name, self.envvars.CONTAINER_REGISTRY_SERVER, branch))
```
adashen: same as above

iotedgedev/modules.py
```diff
@@ -15,6 +14,51 @@ def __init__(self, envvars, utility, output, dock):
...
+            self.output.header(cmd)
+            self.utility.exe_proc(cmd.split(), cwd=self.envvars.MODULES_PATH)
+        elif lang == "csharpfunction":
+            cmd = "dotnet new -i Microsoft.Azure.IoT.Edge.Function"
+            self.output.header(cmd)
+            self.utility.exe_proc(cmd.split())
+            cmd = "dotnet new aziotedgefunction -n {0} -r {1}".format(name, self.envvars.CONTAINER_REGISTRY_SERVER)
```
adashen: same as above

iotedgedev/cli.py
```diff
@@ -423,6 +423,25 @@ def azure(setup,
...
+# @click.option('--create',
+#               default=".",
+#               required=False,
+#               help="Creates a new Azure IoT module. User `--create .` to create in current folder. Use `--create TEXT to create in a subfolder.`")
```
jongio: "Use" not "User"

iotedgedev/modules.py
```diff
@@ -15,6 +14,53 @@ def __init__(self, envvars, utility, output, dock):
         self.dock = dock
         self.dock.init_registry()
 
+    def create(self, name, template):
+        self.output.header("CREATING MODULE")
+
+        repo = "{0}/{1}".format(self.envvars.CONTAINER_REGISTRY_SERVER, name.lower())
```
jongio: Sanitize name to remove invalid dotnetnamespace chars, like "-"

iotedgedev/modules.py
```diff
@@ -15,6 +14,53 @@ def __init__(self, envvars, utility, output, dock):
...
+        cwd = self.envvars.MODULES_PATH
+        if template == "csharp":
```
jongio: This whole block can be refactored.  There's a lot that can be extracted into common modules.  I also don't like the idea of calling exe_proc from this module.  

Can you create a class that wraps "dotnet"?  And then also change other calls to dotnet, like in dotnetmoduleprocessor.py to use that new dotnetcls.py class you create.

iotedgedev/modules.py
```diff
@@ -15,6 +14,53 @@ def __init__(self, envvars, utility, output, dock):
...
+            cmd = "dotnet new -i Microsoft.Azure.IoT.Edge.Module"
```
jongio: We'll need to have a method that verifies dotnet is installed.  There's another PR open for that.

iotedgedev/modules.py
```diff
@@ -15,6 +14,53 @@ def __init__(self, envvars, utility, output, dock):
...
+            self.output.header(cmd)
+            self.utility.exe_proc(cmd.split())
+            cmd = "dotnet new aziotedgemodule -n {0} -r {1}".format(name, repo)
+            self.output.header(cmd)
+            self.utility.exe_proc(cmd.split(), cwd=cwd)
+        elif template == "nodejs":
+            pass
```
jongio: Consider using the node_solution template that is in the test folder.  Let's figure out a way to support node with a template.

iotedgedev/modules.py
```diff
@@ -15,6 +14,53 @@ def __init__(self, envvars, utility, output, dock):
...
+            self.utility.exe_proc(cmd.split(), cwd=cwd)
+        elif template == "csharpfunction":
+            cmd = "dotnet new -i Microsoft.Azure.IoT.Edge.Function"
+            self.output.header(cmd)
+            self.utility.exe_proc(cmd.split())
+            cmd = "dotnet new aziotedgefunction -n {0} -r {1}".format(name, repo)
+            self.output.header(cmd)
+            self.utility.exe_proc(cmd.split(), cwd=cwd)
+
+        new_module = """{
```
jongio: I feel like the below code could be extracted into a new class called deploymentfile.py that has a method called add.

Then you can easily test it.

iotedgedev/modules.py
```diff
@@ -15,6 +14,53 @@ def __init__(self, envvars, utility, output, dock):
         self.dock = dock
         self.dock.init_registry()
 
+    def create(self, name, template):
```
jongio: Please write tests for this new method.

iotedgedev/cli.py
jongio: There are two API surfaces due to legacy reasons. 

`module --create` follows the old pattern

`iotedgedev createmodule` follows the new pattern


See in cli.py how I have both for say iotedgedev build and iotedgedev modules --build, first one calls the second one.

iotedgedev/modules.py
LazarusX: We have created a [Yeoman template](https://www.npmjs.com/package/generator-azure-iot-edge-module) for Node.js module. We would like to announce and document it but in the last minute were informed of the potential breaking changes introduced by edgelet. After it is fully tested on the new runtime, we can integrate it here.

iotedgedev/cli.py
LazarusX: Thanks. I have added the new style one. In addition, the command is renamed from "create module" to "add module" for simplicity and aligning with VS Code extension's wording. I will update README once we finalize the API.

iotedgedev/cli.py
LazarusX: Fixed. Thanks.

iotedgedev/modules.py
LazarusX: Added input check. Thanks.

iotedgedev/modules.py
LazarusX: Thanks. I have added a new `DotNet` module to wrap .NET Core CLI operations. For Node.js and Python module, I don't think it worthwhile to extract the `yo` and `cookiecutter` operations to separate modules since they will be only used when adding new modules (*i.e.*, in this class).

iotedgedev/modules.py
LazarusX: Thanks. Then shall we leave this feature to that PR?

iotedgedev/modules.py
LazarusX: Thanks. I have added a new `DeploymentManifest` module.

iotedgedev/modules.py
LazarusX: Added. Thanks.

tests/test_iotedgedev.py
```diff
@@ -203,3 +235,12 @@ def test_load_no_dotenv():
     # assert '.env.test file not found on disk.' in result.output
     # assert 'PROCESSING' in result.output
 '''
+
+
+def add_module_and_verify(main, runner, template):
```
druttka: I'm currently getting this error during the py27 portion of a tox run. py36 seems to get through it. It might be something wrong on my machine, but please double check.

```
================================== FAILURES ===================================
_______________________________ test_module_add _______________________________

    def test_module_add():
        """Test the addmodule command"""
        os.chdir(test_solution_dir)
        cli = __import__("iotedgedev.cli", fromlist=['main'])
        runner = CliRunner()

>       add_module_and_verify(cli.main, runner, "csharp")

..\test_iotedgedev.py:106:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

main = <click.core.Group object at 0x0000000005C04FD0>
runner = <click.testing.CliRunner object at 0x0000000005C619E8>
template = 'csharp'

    def add_module_and_verify(main, runner, template):
        module_name = template + "module"
        result = runner.invoke(main, ['addmodule', module_name, '--template', template])
        print(result.output)
>       assert 'CREATE COMPLETE' in result.output
E       assert 'CREATE COMPLETE' in '\n=========================================\n======== PROCESSING CONFIG FILES ========\n================================...ed in 858.83 ms for E:\iotedgedev\tests\test_solution\modules\csharpmodule\cshar\nRestore succeeded.\n'
E        +  where '\n=========================================\n======== PROCESSING CONFIG FILES ========\n=============\nRestore succeeded.\n' = <Result IOError(22, "invalid mode ('r') or filename")>.outputarpmodule\csharpmodule.csproj.

..\test_iotedgedev.py:244: AssertionError
```

iotedgedev/envvars.py
```diff
@@ -120,6 +120,7 @@ def load(self, force=False):
                 self.CONFIG_OUTPUT_DIR = self.get_envvar("CONFIG_OUTPUT_DIR", default="config")
                 self.DEPLOYMENT_CONFIG_FILE = self.get_envvar("DEPLOYMENT_CONFIG_FILE", altkeys=['MODULES_CONFIG_FILE'])
                 self.DEPLOYMENT_CONFIG_FILE_PATH = os.path.join(self.CONFIG_OUTPUT_DIR, self.DEPLOYMENT_CONFIG_FILE)
+                self.DEPLOYMENT_CONFIG_TEMPLATE_FILE = self.get_envvar("DEPLOYMENT_CONFIG_TEMPLATE_FILE")
```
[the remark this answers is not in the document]
LazarusX: Thanks. I have given a default value for it. In addition, I also added migration method for legacy users: https://github.com/Azure/iotedgedev/pull/180/files#diff-aeb3b85c90a3accbce23dc060891d24cR19

iotedgedev/modules.py
```diff
@@ -15,6 +16,47 @@ def __init__(self, envvars, utility, output, dock):
...
+            self.output.error("Module \"{0}\" already exists under {1}".format(name, os.path.abspath(self.envvars.MODULES_PATH)))
+            return
+
+        repo = "{0}/{1}".format(self.envvars.CONTAINER_REGISTRY_SERVER, name.lower())
+        if template == "csharp":
+            dotnet = DotNet(self.envvars, self.output, self.utility)
+            dotnet.install_module_template()
+            dotnet.create_custom_module(name, repo, cwd)
+        elif template == "nodejs":
+            cmd = "yo azure-iot-edge-module -n {0} -r {1}".format(name, repo)
```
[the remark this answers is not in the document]
LazarusX: Thanks. Now it will check dependency and fail fast if the dependency is not installed. In addition, I also updated README to add instructions on setting up environment manually https://github.com/Azure/iotedgedev/pull/180/files#diff-04c6e90faac2675aa89e2176d2eec7d8

iotedgedev/modules.py
```diff
@@ -15,6 +16,47 @@ def __init__(self, envvars, utility, output, dock):
         self.dock = dock
         self.dock.init_registry()
 
+    def add(self, name, template):
+        self.output.header("CREATING MODULE")
```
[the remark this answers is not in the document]
LazarusX: Thanks. That's my mistake. I forgot to update all the terminologies. Just fixed them.

tests/test_iotedgedev.py
LazarusX: Can you help check if there is a valid DEPLOYMENT_CONFIG_TEMPLATE_FILE key in your .env file? I also added more detailed error messages in this update. Please help run the test again and let me know if the error still exists.

iotedgedev/utility.py
```diff
@@ -40,6 +40,13 @@ def call_proc(self, params, shell=False, cwd=None):
         except Exception as e:
             self.output.error("Error while executing command: {0}. {1}".format(' '.join(params), str(e)))
 
+    def check_dependency(self, params, description, shell=False):
+        try:
+            self.exe_proc(params, shell=shell)
+        except:
+            self.output.error("The {0} is required by the Azure IoT Edge Dev Tool. For installation instructions, see the README at https://aka.ms/iotedgedev.".format(description))
```
druttka: I like that this is done in a check_dependency utility now, and thanks for updating the instructions to go with it! 

One piece of feedback I got on the original message here is that the message should be clear that {tool} is only required if the user is trying to build {lang/platform} modules. 

Maybe consider dropping "The " from the front, so the format string is:
`"{0} is required by the Azure IoT Edge Dev Tool. For installation instructions, see the README at https://aka.ms/iotedgedev."`

And then the description passed in can be of the form
`"To build python modules, the Cookiecutter tool "`
or
`"To build .NET modules, the .NET Core SDK "`

etc.

tests/test_iotedgedev.py
druttka: The changes to handle this at runtime for the user look very good, and the test is OK if I put DEPLOYMENT_CONFIG_TEMPLATE_FILE in .env.

There is still a test failure when DEPLOYMENT_CONFIG_TEMPLATE_FILE is not present. It seems like `add_module_and_verify` goes directly to `os.environ`, so it doesn't get the default from `envvars`? 

Other than that case, looks good to me.

LazarusX: Thanks for the test. I can reproduce it on my side. I fixed it by setting `os.environ` when `get_envvar` falls back to default value. Please try again.

iotedgedev/utility.py
LazarusX: Thanks for the information! I have adopted your awesome proposal.

tests/test_iotedgedev.py
druttka: This fixes the test, thanks!

I'm not sure that we should mutate `os.environ` in every case that we take a default value, though. I don't have a very strong opinion here. I believe that in python, setting `os.environ` is scoped to the process and child processes, so setting it should not have undesired side effects for the user. 

With that in mind, two alternatives to consider.

One would be to have `test_iotedgedev` import and load `envvars`. Then the code in `add_module_and_verify` can look like 
`    assert module_name in json.load(open(envvars.DEPLOYMENT_CONFIG_TEMPLATE_FILE))["moduleContent"]["$edgeAgent"]["properties.desired"]["modules"]`, 

With this approach, no change to the `envvars` behavior is required. The test would benefit from the existing `load` setting the default value on `self`, and we avoid touching the underlying `os.environ`. 

Since `test_iotedgedev` runs end-to-end tests, I'm comfortable with the direct dependency on `envvars`. However, if we wanted to avoid that direct dependency, a second alternative might be to promote our default value to a constant. Then `test_iotedgedev` can check `os.environ`, but fallback to the constant.

LazarusX: Yes...Actually I considered alternative solutions including the ones you proposed but failed to find a strong reason to prefer one over another so I simply chose the easiest one :) Thanks for your kind suggestions anyway.

PR #181, opened 2018-06-06
add contributing.md and modify readme to point to wiki

kln4dt:


README.md
```diff
@@ -1,977 +1,5 @@
...
-#### Run Tests
-
-Run the following command to run tests.
-    
-`tox`
-    
-    
+The **Azure IoT Edge Dev Tool** greatly simplifies [Azure IoT Edge](https:/azure.microsoft.com/en-us/services/iot-edge/) development.
 
+Please refer to the [Wiki](https://github.com/Azure/iotedgedev/wiki) for details on setup, usage, and troubleshooting.
```
jongio: Let's have the quickstart duplicated here, so it's not just a blank page.

Also include a link to the contributing page.

Also, from the wiki, have a contributing node in the TOC, but point it to the contributing.md doc.


You get some nice stuff on github if you have the contributing.md file https://blog.github.com/2012-09-17-contributing-guidelines/

kln4dt:
Will update the TOC once this is merged to avoid pointing to a nonexistent file

README.md
```diff
@@ -1,5 +1,119 @@
...
 
-The **Azure IoT Edge Dev Tool** greatly simplifies [Azure IoT Edge](https:/azure.microsoft.com/en-us/services/iot-edge/) development.
+The **Azure IoT Edge Dev Tool** greatly simplifies [Azure IoT Edge](https:/azure.microsoft.com/en-us/services/iot-edge/) development down to simple commands driven by Environment Variables. 
 
-Please refer to the [Wiki](https://github.com/Azure/iotedgedev/wiki) for details on setup, usage, and troubleshooting.
\ No newline at end of file
+ - It gets you started with IoT Edge development with the [IoT Edge Dev Container](#iot-edge-dev-container) and IoT Edge Solution Scaffolding that contains a sample module and all the required configuration files.
+ - It speeds up your inner-loop dev (dev, debug, test) by reducing multi-step build & deploy processes into one-line CLI commands and well as drive your outer-loop CI/CD pipeline. _You can use all the same commands in both stages of your development life-cycle._
+
+## Resources
```
jongio: Please put this on bottom of page

README.md
```diff
@@ -1,5 +1,119 @@
...
+
+1. **View Logs**: View and Save Docker log files: 
+
+    `iotedgedev docker --logs`
+
+1. **Setup Custom Registry**: Use a Custom Container Registry: 
+
+    `iotedgedev docker --setup-registry`
+
+Please see [Azure IoT Edge Dev Resources](https://github.com/jonbgallant/azure-iot-edge-dev) for links to official docs and other IoT Edge dev information.
```
jongio: I think this line can be removed.

README.md
```diff
@@ -1,5 +1,119 @@
...
+    
+1. **Start the IoT Edge Runtime**
+
+    `iotedgedev start`
+
+1. **Monitor Messages sent from IoT Edge to IoT Hub**
+
+    `iotedgedev monitor`
+
+## Overview
```
jongio: Let's only do the quickstart, not the overview.

README.md
```diff
@@ -63,57 +58,7 @@ The only thing you need to install is Docker. All of the other dev dependencies
...
-    `iotedgedev monitor`
-
-1. **View Logs**: View and Save Docker log files: 
-
-    `iotedgedev docker --logs`
-
-1. **Setup Custom Registry**: Use a Custom Container Registry: 
-
-    `iotedgedev docker --setup-registry`
+## Resources
```
jongio: When looking at the output html file, I feel like we are using the reader hanging at the end of the Quickstart.

Can you come up with a good way to say something like CLICK HERE TO VIEW ALL DOCS, not that exactly, but something like that.

README.md
```diff
@@ -4,11 +4,6 @@ The **Azure IoT Edge Dev Tool** greatly simplifies [Azure IoT Edge](https:/azure
 
  - It gets you started with IoT Edge development with the [IoT Edge Dev Container](#iot-edge-dev-container) and IoT Edge Solution Scaffolding that contains a sample module and all the required configuration files.
  - It speeds up your inner-loop dev (dev, debug, test) by reducing multi-step build & deploy processes into one-line CLI commands and well as drive your outer-loop CI/CD pipeline. _You can use all the same commands in both stages of your development life-cycle._
```
jongio: I feel like we need a link to the wiki at the top, maybe after the intro.  Something like....for full docs see the wiki, link.

kln4dt: Agreed, that's what I initially had but then coupled it with contributing.md in resources. Want to split that back out or leave the link there and add another reference at the top?

jongio: Let's add another at top...just not under resources header...probably in overview section would be best.

README.md
kln4dt: which links to the wiki? in that case, think it would make sense to pull out the reference to the wiki at the beginning of the resources section and have it here

jongio: How do you feel about leaving as-is for now and revisit it later?

kln4dt: yeah, will be good to re-read it with some fresher eyes

jongio: Okay, let's still do the link at the top, comment below and call it good.

PR #188, opened 2018-06-14
add note to monitor events call that it could take some time

kln4dt:


iotedgedev/iothub.py
```diff
@@ -15,6 +15,8 @@ def monitor_events(self, timeout=0):
             timeout = 0
 
         try:
+            self.output.status("Monitoring events may take longer than expected...")
```
druttka: It's a great idea to add the note. What do you think about adding it to iothub-explorer instead? 

Maybe changing the iothub-explorer "Monitoring events from device {id}..." output to something like "Connected to device {id}. If no events are observed, check for common solutions at {aka.ms link to be created}."  The aka.ms could have some common FAQs like how long can it take, how consumer groups might affect the monitoring, etc.

jongio: iothub-explorer is being replaced by new IoT CLI extension. @digimaun is working on it. Maybe he can figure out a message to display for monitor-events?  Let's not update iothub-explorer.

jongio: @kln4dt Let's go with "It may take 1-2 minutes before you start to see messages below."  It's more about that it will take time, but not that there's an expectation for it.

druttka: I like that wording. 

Fair enough not to update iothub-explorer. Sounds good for @digimaun to consider whether a similar note in the CLI ext would be useful for non-iotedgedev users. What do you think about the message including a link about why they might not be seeing anything come through?

jongio: @druttka  - good idea, but that doc doesn't exist - so i think the answer is finish this, create task to create the doc and then update this.

druttka: Agree. In case it wasn't clear, my Approve on this change is independent of the ideas for the future.

PR #193, opened 2018-06-26
Support for multiple registries 

kln4dt:
Ongoing PR
Done:
creation of new container registry class
parsing of .env file for multiple registries
refactored modules.py to use module.json for container registry value

To Do:
refactor dockercls.py
testing


iotedgedev/envvars.py
```diff
@@ -209,6 +207,25 @@ def save_envvar(self, key, value):
...
+        self.CONTAINER_REGISTRY = {}
+        for key in os.environ:
+            if key.startswith('CONTAINER_REGISTRY_SERVER'):
+                token = key[25:]
+                if token not in registries:
+                    registries[token] = {}
+                registries[token]['server'] = os.environ[key]
+            elif key.startswith(('CONTAINER_REGISTRY_USERNAME', 'CONTAINER_REGISTRY_PASSWORD')):
+                token = key[27:]
+                value = key[19:27].lower()
```
druttka: Not a big deal, but the name "value" here tripped up my brain for a second. I'm not sure what the best alternative name is, though. Maybe "subkey"?

iotedgedev/envvars.py
```diff
@@ -209,6 +207,25 @@ def save_envvar(self, key, value):
...
+                if token not in registries:
+                    registries[token] = {}
+                registries[token][value] = os.environ[key]
+
+        for key, value in registries.items():
+            self.CONTAINER_REGISTRY[key] = ContainerRegistry(value['server'], value['username'], value['password'])
```
druttka: Looks good. Do we need to assign the old properties for any back-compatibility or will all usages be internal, and dig into this for the registry w/default token? I'm thinking
```
defaultRegistry = registries['']
self.CONTAINER_REGISTRY_SERVER = defaultRegistry['server']
...
```

iotedgedev/envvars.py
```diff
@@ -74,7 +75,7 @@ def load(self, force=False):
                 self.output.header("ENVIRONMENT VARIABLES")
 
             self.load_dotenv()
-
+            self.get_registries()
```
jongio: This  should be in the try so we don't hose on load or parse.

iotedgedev/envvars.py
```diff
@@ -209,6 +207,25 @@ def save_envvar(self, key, value):
             self.output.error(f("Could not update the environment variable {key} in file {dotenv_path}"))
             sys.exit(-1)
 
+    def get_registries(self):
+        registries = {}
+        self.CONTAINER_REGISTRY = {}
+        for key in os.environ:
+            if key.startswith('CONTAINER_REGISTRY_SERVER'):
```
jongio: Is this a case sensitive search?

iotedgedev/modules.py
```diff
@@ -128,9 +128,18 @@ def build_push(self, no_build=False, no_push=False):
                         if not no_push:
                             # PUSH TO CONTAINER REGISTRY
                             self.output.info("PUSHING DOCKER IMAGE TO: " + image_destination_name)
+                            # FIND ENV VAR KEY BASED ON REGISTRY VALUE
+                            registry_key = None
+                            for registry in self.envvars.CONTAINER_REGISTRY:
```
jongio: Should this look up be in the ContainerRegistry class?

iotedgedev/modules.py
```diff
@@ -128,9 +128,18 @@ def build_push(self, no_build=False, no_push=False):
...
+                                if self.envvars.CONTAINER_REGISTRY[registry].server == module_json.registry:
+                                    registry_key = registry
+                                    break
+                            if registry_key is None:
+                                self.output.error("Error while trying to retrieve container registry information.")
```
jongio: How about "Could not find registry server with key {0}. Please make sure your envvar is set.

iotedgedev/modules.py
jongio: Then you would just to a cr = ContainerRegistry.get(key)

and then later.

cr.Server
cr.Password
cr.Username

jongio:
@LazarusX - Can you please keep an eye on this PR and review?  THanks, Jon

iotedgedev/envvars.py
```diff
@@ -209,6 +209,26 @@ def save_envvar(self, key, value):
...
+        for key in os.environ:
+            key = key.lower()
+            if key.startswith('container_registry_server'):
+                token = key[25:]
+                if token not in registries:
+                    registries[token] = {}
+                registries[token]['server'] = os.environ[key]
+            elif key.startswith(('container_registry_username', 'container_registry_password')):
+                token = key[27:]
+                subkey = key[19:27]
```
LazarusX: There are a few magic numbers in this method (25, 27, 19, 27). Can you use variable or code (such as len(`'container_registry_server'`) to explain them?

LazarusX:
When building the solution, we should also update the `runtime.settings.registryCredentials
.{registryId}.username` section (https://docs.microsoft.com/en-us/azure/iot-edge/module-edgeagent-edgehub#edgeagent-desired-properties) to add required Docker credentials in the solution.

LazarusX:
@jongio @kln4dt  As we will support multiple registries, we might require users to input repositories, registry username and password when adding modules. The registry and credential information will then be saved in env.

iotedgedev/envvars.py
```diff
@@ -215,13 +215,13 @@ def get_registries(self):
         for key in os.environ:
             key = key.lower()
             if key.startswith('container_registry_server'):
-                token = key[25:]
+                token = key[len('container_registry_server'):]
```
jongio: I would abstract 'container_registry_server' into a const, so it is not repeated in code. If you need to change later you change in one place.

.env.tmp
```diff
@@ -18,6 +18,13 @@ CONTAINER_REGISTRY_SERVER="localhost:5000"
 CONTAINER_REGISTRY_USERNAME=""
 CONTAINER_REGISTRY_PASSWORD=""
 
+    # To specify additional container registries ensure the prefix is CONTAINER_REGISTRY_SERVER, CONTAINER_REGISTRY_USERNAME, CONTAINER_REGISTRY_PASSWORD
+    # And the token following the prefix uniquely associates the SERVER/USERNAME/PASSWORD
```
druttka: Very minor, but can we explicitly call out for the reader whether the token must be numeric, versus a string like "CORPORATE"?

iotedgedev/envvars.py
```diff
@@ -209,6 +209,26 @@ def save_envvar(self, key, value):
             self.output.error(f("Could not update the environment variable {key} in file {dotenv_path}"))
             sys.exit(-1)
 
+    def get_registries(self):
+        registries = {}
+        self.CONTAINER_REGISTRY = {}
```
druttka: Extremely minor, consider renaming self.CONTAINER_REGISTRY to something like CONTAINER_REGISTRY_MAP or CONTAINER_REGISTRIES_BY_SERVER, etc. Current name could be interpreted as an object describing a single registry.

iotedgedev/envvars.py
```diff
@@ -209,6 +209,26 @@ def save_envvar(self, key, value):
...
+        for key in os.environ:
+            key = key.lower()
+            if key.startswith('container_registry_server'):
+                token = key[len('container_registry_server'):]
+                if token not in registries:
+                    registries[token] = {}
+                registries[token]['server'] = os.environ[key]
+            elif key.startswith(('container_registry_username', 'container_registry_password')):
+                token = key[len('container_registry_username'):]
```
druttka: Consider defining a var like (sorry about the long name) KEY_LENGTH_FOR_USERNAME_OR_PASSWORD. This would create some self-describing code for future readers/maintainers. Right now, they might have a brief moment of doubt before tying together that username and password are the same length.

iotedgedev/modules.py
```diff
@@ -89,8 +89,8 @@ def build_push(self, no_build=False, no_push=False):
             for platform in module_json.platforms:
                 # get the Dockerfile from module.json
                 dockerfile = os.path.abspath(os.path.join(module_dir, module_json.get_dockerfile_by_platform(platform)))
-                container_tag = "" if self.envvars.CONTAINER_TAG == "" else "-" + self.envvars.CONTAINER_TAG
-                tag = "{0}:{1}{2}-{3}".format(module_json.repository, module_json.tag_version, container_tag, platform).lower()
+                #container_tag = "" if self.envvars.CONTAINER_TAG == "" else "-" + self.envvars.CONTAINER_TAG
+                tag = "{0}:{1}{2}-{3}".format(module_json.repository, module_json.tag_version, "", platform).lower()
```
druttka: I'm probably missing context on this change, but it looks like the new behavior is to always use empty string where we used to potentially use "-{container_tag}". In that case, should we remove the placeholder from the format altogether, and end up with the following?

`tag = "{0}:{1}-{2}".format(module_json.repository, module_json.tag_version, platform).lower()`

jongio: CONTAINER_TAG is EnvVar that allows you to add additional bits to the tag to make it unique.  That shouldn't be commented out.

kln4dt: Accidentally left this commented while testing. Will uncomment

jongio:
@kln4dt @LazarusX  - Can you take the deployment.json updates in a diff issue so we can get this pushed to unlock other issues?  Thanks, Jon

LazarusX:
@jongio I am fine with that. Filed a separate issue https://github.com/Azure/iotedgedev/issues/228 for tracking.

iotedgedev/envvars.py
```diff
@@ -234,6 +234,29 @@ def save_envvar(self, key, value):
             self.output.error(f("Could not update the environment variable {key} in file {dotenv_path}"))
             sys.exit(-1)
 
+    def get_registries(self):
+        registries = {}
+        self.CONTAINER_REGISTRY_MAP = {}
+        length_container_registry_server = len('container_registry_server')
+        length_container_registry_username_or_password = len('container_registry_username')
+        length_container_registry_ = len('container_registry_')
```
jongio: Please don't end variable name with an underscore.

iotedgedev/modules.py
```diff
@@ -138,9 +138,17 @@ def build_push(self, no_build=False, no_push=False):
...
+                        if registry.server.lower() == tag.split('/')[0].lower():
+                            registry_key = key
+                            break
+                    if registry_key is None:
+                        self.output.error("Could not find registry server with name {0}. Please make sure your envvar is set.".format(tag.split('/')[0].lower()))
+                    self.output.info("module json reading {0}".format(tag))
 
                     for line in self.dock.docker_client.images.push(repository=tag, stream=True, auth_config={
-                            "username": self.envvars.CONTAINER_REGISTRY_USERNAME, "password": self.envvars.CONTAINER_REGISTRY_PASSWORD}):
+                            "username": self.envvars.CONTAINER_REGISTRY_MAP[registry_key].username, "password": self.envvars.CONTAINER_REGISTRY_MAP[registry_key].password}):
```
jongio: Since you are doing hte lookup twice, you should get the object once and then use the props on that object.
```
cr = self.envvars.CONTAINER_REGISTRY_MAP[key]
cr.username
cr.password
```

iotedgedev/modules.py
```diff
@@ -138,9 +138,17 @@ def build_push(self, no_build=False, no_push=False):
                 if not no_push:
                     # PUSH TO CONTAINER REGISTRY
                     self.output.info("PUSHING DOCKER IMAGE: " + tag)
+                    registry_key = None
+                    for key, registry in self.envvars.CONTAINER_REGISTRY_MAP.items():
+                        if registry.server.lower() == tag.split('/')[0].lower():
```
jongio: Please add a sample of the string you are parsing here in a comment, so it's easier to understand the code.

iotedgedev/envvars.py
```diff
@@ -234,6 +234,29 @@ def save_envvar(self, key, value):
             self.output.error(f("Could not update the environment variable {key} in file {dotenv_path}"))
             sys.exit(-1)
 
+    def get_registries(self):
```
jongio: Comments in here might help understand what this is doing.

tests/test_envvars.py
```diff
@@ -148,3 +146,28 @@ def test_is_terse_command_empty():
     output = Output()
     envvars = EnvVars(output)
     assert envvars.is_terse_command("")
+
+def test_default_container_registry_server_key_exists():
```
jongio: Add some tests for a custom (non-default container registry server).

iotedgedev/dockercls.py
```diff
@@ -66,6 +68,8 @@ def init_local_registry(self):
             self.output.info("Running registry container")
             self.docker_client.containers.run("registry:2", detach=True, name="registry", ports=ports, restart_policy={"Name": "always"})
 
+<<<<<<< HEAD
```
LazarusX: Unresolved conflict?

jongio:
Seeing this in Travis log
```
.env file not found on disk. Without a file on disk, you must specify all Environment Variables at the system level. (/home/travis/build/Azure/iotedgedev/.env)
ERROR: Environment variables not configured correctly. Run `iotedgedev solution create` to create a new solution with sample .env file. Please see README for variable configuration options. Tip: You might just need to restart your command prompt to refresh your Environment Variables.
ERROR: Variable that caused exception: 'username'

```

@druttka - Did you config Travis with a .env?
@kln4dt - Why is it looking for username variable in envvar?

kln4dt:
@jongio @druttka 
The 'username' issue has been fixed as we discussed yesterday, not making the assumption that the username or password will be specified in .env if auth isn't required.
I pushed new changes reflecting that business logic + additional tests + fixes based on comments.

2 of my unit tests are failing - why the travis build is failing. Related to the .env file that it's testing against. Hoping one of you can provide more clarity since it's not finding one of the .env variables in the file.

Issue #170, opened 2018-05-08
Add lang param to iotedgedev solution


jongio:
iotedgedev solution [solutionname] --lang [lang|c#]

Update iotedgedev to accept a lang param, default to C#

PR #194, opened 2018-06-28
Add --template param to solution creation

LazarusX:
This PR implements #170. Now creating a solution, you must specify the default module name and template, e.g., `iotedgedev solution test_solution --module filtermodule --template csharp`.

Breaking changes in this PR: now it will parse deployment.template.json to find modules to process instead of respecting the `ACTIVE_MODULES` env. @jongio Please tell me if you have any concern or suggestion.

Due to a issue with GA version of EdgeHub when working with iotedgectl, the test will fail. We have reported this to IoT Edge team.


iotedgedev/modules.py
```diff
@@ -66,72 +65,71 @@ def push(self, no_build=False):
...
-
-            if len(modules_to_process) == 0 or modules_to_process[0] == "*" or module in modules_to_process:
-
-                module_dir = os.path.join(self.envvars.MODULES_PATH, module)
-
-                self.output.info("BUILDING MODULE: {0}".format(module_dir), suppress=no_build)
-
-                module_json = Module(self.output, self.utility, os.path.join(module_dir, "module.json"))
-                mod_proc = ModulesProcessorFactory(self.envvars, self.utility, self.output, module_dir).get(module_json.language)
+        deployment_manifest = DeploymentManifest(self.envvars, self.output, self.envvars.DEPLOYMENT_CONFIG_TEMPLATE_FILE, True)
```
LazarusX: Breaking changes here: now it will parse deployment.template.json to find modules to process instead of respecting the `ACTIVE_MODULES` env. @jongio Please tell me if you have any concern or suggestion.

iotedgedev/cli.py
```diff
@@ -61,14 +61,22 @@ def main(set_config, az_cli=None):
               required=False,
               help="Creates a new Azure IoT Edge Solution. Use `--create .` to create in current folder. Use `--create TEXT` to create in a subfolder.")
 @click.argument("name", required=False)
-def solution(create, name):
+@click.option('--module',
```
jongio: I would prefer to have a default module name.  Let's think of something that works well. I used filtermodule before, but can be anything.

iotedgedev/deploymentmanifest.py
```diff
@@ -37,13 +37,37 @@ def add_module_template(self, module_name):
...
         self.json["moduleContent"]["$edgeAgent"]["properties.desired"]["modules"][module_name] = json.loads(new_module)
 
+        self.add_default_route(module_name)
+
+    def add_default_route(self, module_name):
+        """Add a default route to send messages to IoT Hub"""
+        new_route_name = "{0}ToIoTHub".format(module_name)
+        new_route = "FROM /messages/modules/{0}/outputs/* INTO $upstream".format(module_name)
+
+        self.json["moduleContent"]["$edgeHub"]["properties.desired"]["routes"][new_route_name] = new_route
```
jongio: Seems like we should check if the values exist.  Thoughts on loaded the deployment file into memory object instead of json, and then serialize to json when done?  I think the interface to the deployment manifest would be simplified that way.

iotedgedev/cli.py
LazarusX: Shall we add a new environment entry like `DEFAULT_MODULE_NAME` which defaults to "filtermodule"?

iotedgedev/deploymentmanifest.py
LazarusX: @jongio I added a `nested_set` utility method to ensure that the keys exist when we set the values in a nested way.

iotedgedev/cli.py
jongio: Sure

iotedgedev/utility.py
```diff
@@ -112,7 +112,7 @@ def get_modules_in_config(self, moduleType):
             return_modules.update(user_modules)
             return return_modules
 
-    def set_config(self, force=False):
+    def set_config(self, force=False, var_dict=None):
```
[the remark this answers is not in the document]
LazarusX: @jongio How about `key_value_map`? Do you have any suggestions?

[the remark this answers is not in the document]
jongio: replacements?

jongio:
@LazarusX - Can you think about what tests can be added to this to cover your newly added code?

iotedgedev/cli.py
LazarusX: I just realize that the environment variables loaded by .env file are scoped to a solution. Before the user runs the `iotedgedev solution --create` command, the .env file is not created yet. So the user will need to set the environment variable by himself. My current design is in https://github.com/Azure/iotedgedev/pull/194/files#diff-f7f95b5a237aedfa9c80a3ec57d525e7R66. Please let me know if you have better ideas. :)

jongio: IC.  Yes, hardcode the default name, allow overrides via ENV or CMD line.

iotedgedev/envvars.py
```diff
@@ -109,7 +109,7 @@ def load(self, force=False):
                 self.RUNTIME_CONFIG_DIR = self.get_envvar("RUNTIME_CONFIG_DIR", default=".")
                 if self.RUNTIME_CONFIG_DIR == ".":
                     self.set_envvar("RUNTIME_CONFIG_DIR", self.get_runtime_config_dir())
-                self.ACTIVE_MODULES = self.get_envvar("ACTIVE_MODULES")
```
jongio: This is a breaking change, so we have to make sure this is documented in the release notes.  Please make a note somewhere or file an issue to track these for the next release.  THanks,

iotedgedev/deploymentmanifest.py
```diff
@@ -10,12 +10,13 @@
...
 class DeploymentManifest:
-    def __init__(self, envvars, output, path, is_template):
+    def __init__(self, envvars, output, utility, path, is_template):
+        self.utility = utility
         self.output = output
         try:
             self.path = path
             self.is_template = is_template
-            self.json = json.load(open(path))
+            self.json = json.loads(os.path.expandvars(self.utility.get_file_contents(path)))
```
jongio: Let's put this expand/get command in utility.

iotedgedev/module.py
```diff
@@ -18,7 +18,7 @@ def load_module_json(self):
         if os.path.exists(self.module_json_file):
             try:
                 self.file_json_content = json.loads(
-                    self.utility.get_file_contents(self.module_json_file))
+                    os.path.expandvars(self.utility.get_file_contents(self.module_json_file)))
```
jongio: Another use of expand/get file to put in utility.

iotedgedev/deploymentmanifest.py
LazarusX: Moved. Thanks.

iotedgedev/envvars.py
LazarusX: Sure. Issue filed: https://github.com/Azure/iotedgedev/issues/201

LazarusX:
@jongio, below are major changes since your last review:
* Add update `get_file_contents` method in utility to include `os.expandvars`
* Fix a issue with Docker for Windows with Linux mode
* Update requirements.txt to update Docker SDK version (while includes a necessary bug fix) and remove azure-iot-edge-runtime-ctl
* Disable outdated test cases (those based on public preview runtime) temporarily
* Add unit tests for new methods in utility.py and deploymentmanifest.py

Please take a look.

For e2e tests, especially for the `BYPASS_MODULES` support, I am considering opening a separate PR after iotedgehubdev integration because:
* This PR is already getting ridiculously large
* We cannot perform e2e test now using public preview runtime

What do you think?

jongio:
I like that idea.  Yes, we need to replace iotedgectl with a combo of iotedge and iotedgehubdev.  Can you spec that out?  It would like to keep same iotedgedev command line API, but change underlying implementation.

iotedgedev/template/deployment.template.json
```diff
@@ -14,7 +14,7 @@
           "edgeAgent": {
             "type": "docker",
             "settings": {
-              "image": "microsoft/azureiotedge-agent:${RUNTIME_TAG}",
+              "image": "mcr.microsoft.com/azureiotedge-agent:${RUNTIME_TAG}",
```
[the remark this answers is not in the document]
LazarusX: Can you elaborate more on this? I assume the `mcr.microsoft.com/azureiotedge-agent` part is unlikely to change in the future, and they are only used in the two template files.

iotedgedev/utility.py
```diff
@@ -80,9 +80,13 @@ def get_iot_hub_sas_token(self, uri, key, policy_name, expiry=3600):
 
         return "SharedAccessSignature " + urlencode(rawtoken)
 
-    def get_file_contents(self, file):
+    def get_file_contents(self, file, expand_env=False):
```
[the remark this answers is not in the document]
LazarusX: I'm sorry but I don't quite get your point. Did you mean updating the variable name from `expand_env` to `expandvars`?

iotedgedev/template/.env.tmp
```diff
@@ -55,9 +55,10 @@ BYPASS_MODULES=""
     # "" - to build all modules
     # "filtermodule, module1" - Comma delimited list of modules to bypass when building
 
-ACTIVE_DOCKER_PLATFORMS="amd64"
-    # "*" - to build all docker files
-    # "amd64,amd64.debug" - Comma delimted list of docker files to build
+ACTIVE_DOCKER_PLATFORMS=""
```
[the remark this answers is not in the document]
LazarusX: The deployment.template.json will be expanded to deployment.json which will be used for the real deployment. If the user targets the solution for a specific platform, he will always make sure that deployment.template.json reflects his intention. For additional platforms, he can update `ACTIVE_DOCKER_PLATFORMS` to include them. That is my assumption for typical usages.

On the other hand, I want to make this tool works as expected out-of-box. We have many existing documentations introducing the usage of deployment.template.json. Say a user targets the solution to platforms except "amd64" such as "windows-amd64". He knows he needs to update the deployement.template.json. If the `ACTIVE_DOCKER_PLATFORMS` variable contains "amd64", he may encounter build errors, which can be confusing. That's my reason of setting its default value as `""`.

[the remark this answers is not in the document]
LazarusX: In some cases, users may want to build a single more for a specific platform. Maybe we can add a new command like `iotedgedev buildmodule --name filtermodule --platform arm32v7`?

LazarusX:
@jongio Here is the draft spec for the iotedge and iotedgehubdev integration: https://github.com/Azure/iotedgedev/issues/202. Please help review it. :)

There are some ricks I can expect on the iotedge part. Can you share your team's investigation for the iotedge system service integration if any?

Our CI/CD plugins for VSTS and Jenkins will take dependencies on iotedgedev, but some issues such as #198 and #200 currently are blocking. We plan to include the fixes in this release. We also would like to perform some manual e2e tests covering multiple OS and Python 2/3 besides automatic e2e test, since there are so much changes in this release. I think the safest goal for the release should be around middle August.

iotedgedev/template/.env.tmp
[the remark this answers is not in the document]
jongio: Sure, Can you create a task to discuss.?

iotedgedev/template/deployment.template.json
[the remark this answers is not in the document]
jongio: Okay, that's fine.  If it changes for some reason, we can deal with it then. I agree, unlikely.

iotedgedev/utility.py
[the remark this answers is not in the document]
jongio: Yes, but not holding this up for that change.  If you want to change the name, then feel free to create a new task.

iotedgedev/template/.env.tmp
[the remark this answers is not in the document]
LazarusX: https://github.com/Azure/iotedgedev/issues/203 Issue filed. Let's move the discussion there.

iotedgedev/utility.py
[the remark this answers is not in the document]
LazarusX: Sure. Issue filed: https://github.com/Azure/iotedgedev/issues/204

PR #195, opened 2018-06-28
Configure travis for unit tests

druttka:


.travis.yml
```diff
@@ -1,29 +1,12 @@
...
 language: python
 python:
-  - 3.5
-  - 3.4
-  - 3.3
-  - 2.7
-  - 2.6
-
-# command to install dependencies, e.g. pip install -r requirements.txt --use-mirrors
-install: pip install -U tox-travis
```
druttka: Travis already manages a variety of virtual environments for us, so we don't need tox to do that again on top of those. https://docs.travis-ci.com/user/languages/python/#Using-Tox-as-the-Build-Script

.travis.yml
```diff
@@ -1,29 +1,12 @@
...
+  - "3.6"
+  - "3.5"
+  - "3.4"
+  - "3.3"
+  - "2.7"
+  - "2.6"
+install:
+  - pip install -r requirements_travis.txt
+script:
+  - pytest -m unit # & pylint iotedgedev # or py.test for Python versions 3.5 and below
```
druttka: We won't lint yet, but `& pylint iotedgedev` is how we can. See build history for failed builds showing how many lint issues we'd currently have.

tests/__init__.py
```diff
@@ -0,0 +1,7 @@
+# -*- coding: utf-8 -*-
```
druttka: Adding this may have been unnecessary - it looked like a previous issue where files weren't being collected due to absence of __init__.py. I will try removing it.

tests/test_connectionstring.py
```diff
@@ -56,23 +58,4 @@ def test_invalid_devicehub_connectionstring():
     assert connectionstring.HostName == "testhub.azure-devices.net"
     assert connectionstring.HubName == "testhub"
     assert not connectionstring.DeviceId
-    assert connectionstring.SharedAccessKey == "othergibberish"
-
-def test_valid_env_iothub_connectionstring():
```
druttka: These two were moved to the e2e test for lack of a better place. Other option is to mark them as not being unit tests. The underlying issue was that they fail if the .env isn't set, and in a PR validation flow, it won't be.

.travis.yml
```diff
@@ -1,29 +1,10 @@
...
-  user: jonbgallant
-  password:
-    secure: PLEASE_REPLACE_ME
-  on:
-    tags: true
-    repo: azure/iotedgedev
-    python: 2.7
+  - "3.6"
+  - "3.5"
+  - "3.4"
```
druttka: Dropped 3.3 and 2.6 for now due to some [issues](https://travis-ci.org/Azure/iotedgedev/builds/398019345). Tox was only running on 2.7 and 3.6, so we're still getting more than we were. LMK if concerned and we can dig in.

tests/__init__.py
druttka: Seems necessary. [Failed](https://travis-ci.org/Azure/iotedgedev/builds/398020875) on every virtual env when I tried removing it again. Python isn't my usual world, so very possible I'm missing something else. If so, let me know.

.travis.yml
```diff
@@ -1,29 +1,10 @@
...
+  - "2.7"
+install:
+  - pip install -r requirements_travis.txt
+script:
+  - pytest -m unit # & pylint iotedgedev # or py.test for Python versions 3.5 and below
```
druttka: pylint runs successfully, but I'm not including it in the check yet because we have [significant lint failures](https://travis-ci.org/Azure/iotedgedev/jobs/398015885) at the moment.

README.md
```diff
@@ -1,5 +1,7 @@
 # Azure IoT Edge Dev Tool
 
+[![Build Status](https://travis-ci.org/Azure/iotedgedev.svg?branch=master)](https://travis-ci.org/Azure/iotedgedev)
```
druttka: This is just to have the fancy badge on the README if we do indeed use Travis. =)

Issue #219, opened 2018-08-01
Redesign CLI UX


LazarusX:
This is a proposal to redesign the iotedgedev CLI UX to improve consistency and scalability.

### The current status
Currently, there are three CLI styles in iotedgedev:
* **iotedgedev *verb* --*parameter***, for example, `iotedgedev monitor --timeout 100`
* **iotedgedev *noun* --*verb* --*parameter***, for example, `iotedgedev modules --add filtermodule --template csharp`
* **iotedgedev --*verb***, for example, `iotedgedev --set-config`

The problems with these 3 styles are:
* They are just inconsistent.
* For the second style, users would see the verb shown together with the parameter in help messages, for example:
![image](https://user-images.githubusercontent.com/716650/43530980-6097bb68-95e1-11e8-96f6-15db9cf8d1dc.png)
This is little bit confusing since the verb (`add`, `build`, `push`, `deploy`) should be at the same level as the command (`modules`) instead of the parameters (`template`, `no-build`). In addition, most of the parameters only work with specific verb. For example, the `--template` parameter only works when `--add` is specified. If we add more operations to a command in the future, users would see more irrelevant parameters.

### My proposal
Inspired by the scheme of Docker CLI, I would like to propose a new one for iotedgedev. All commands are organized into five categories: `docker`, `iothub`, `runtime`, `simulator` (for iotedgehubdev commands), and `solution`. The general style is **iotedgedev _category_ _command_ --_parameter_**. For some frequently used commands, there is also an **iotedgedev _command_ --_parameter_** style shortcut (only if the shortcut won't cause confusion, similar to the relationship between `docker logs` and `docker container logs`). The shortcuts also add backward compatibility with some current style commands.

Here is what you would see when you type `iotedgedev --help` (shortcuts are intentionally listed before the categories):
![image](https://user-images.githubusercontent.com/716650/43531276-05dddee0-95e2-11e8-924c-2d267fa48e4d.png)

Here is the help message for a category:
![image](https://user-images.githubusercontent.com/716650/43531829-4639af04-95e3-11e8-9d7b-a89cbcf14151.png)

Here is the help message for a specific command. Only the parameters relevant to the command is listed. I also add parameter shortcuts, like `-t` below for `--template`, whenever possible.
![image](https://user-images.githubusercontent.com/716650/43531411-47f65186-95e2-11e8-9961-ed8d2c9fd32e.png)

Below is a full table of the current commands, my proposal, and shortcut (if any):

[...]

Issue #201, opened 2018-07-18
Document breaking changes in GA release


LazarusX:
Document below breaking changes in GA release in CHANGELOG, README, and Wiki:
* `ACTIVE_MODULES` replaced by `BYPASS_MODULES`
* Local dotnet build replaced by container build
* CLI UX redesign (https://github.com/Azure/iotedgedev/issues/219)
* Removal of azure-iot-edge-ctl commands and addition of iotedgehubdev integration
* Requirement of an updated azure-iot-cli-extension

LazarusX:
WIP draft on my fork: https://github.com/LazarusX/iotedgedev/wiki/migration-guides

PR #221, opened 2018-08-02
Redesign CLI UX

LazarusX:
Discussion issue: https://github.com/Azure/iotedgedev/issues/219

All feedbacks are welcome. :)

### DIFF
--- a/iotedgedev/azurecli.py
+++ b/iotedgedev/azurecli.py
@@ -45,7 +45,7 @@ class AzureCli:
             if stderr_data and b"invalid_grant" in stderr_data:
                 self.output.error(self.decode(stderr_data))
                 self.output.info(
-                    "Your Azure CLI session has expired. Please re-run iotedgedev iothub --setup to refresh your credentials.")
+                    "Your Azure CLI session has expired. Please re-run `iotedgedev iothub setup` to refresh your credentials.")
                 self.logout()
                 sys.exit()
 
--- a/iotedgedev/azurecli.py
+++ b/iotedgedev/azurecli.py
@@ -7,8 +7,10 @@ from azure.cli.core import get_default_cli
 from fstrings import f
 output_io_cls = StringIO
 
+
 def get_query_argument_for_id_and_name(token):
     return "[?starts_with(@.id,'{0}') || contains(@.name,'{1}')]".format(token.lower(), token)
+
 
 class AzureCli:
     def __init__(self,  output, envvars, cli=get_default_cli()):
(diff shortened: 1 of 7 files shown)


iotedgedev/azurecli.py
```diff
@@ -43,7 +45,7 @@ def invoke_az_cli_outproc(self, args, error_message=None, stdout_io=None, stderr
             if stderr_data and b"invalid_grant" in stderr_data:
                 self.output.error(self.decode(stderr_data))
                 self.output.info(
-                    "Your Azure CLI session has expired. Please re-run iotedgedev azure --setup to refresh your credentials.")
+                    "Your Azure CLI session has expired. Please re-run `iotedgedev iothub setup` to refresh your credentials.")
```
jongio: This is 'azure' because I want to expand this to include container registry at some point.

iotedgedev/organizedgroup.py
```diff
@@ -0,0 +1,48 @@
+import click
+
+
+class OrganizedGroup(click.Group):
```
jongio: Nice!

iotedgedev/solution.py
```diff
@@ -14,7 +14,7 @@ def create(self, name, module, template):
             dir_path = os.path.join(os.getcwd(), name)
 
         if not self.utility.is_dir_empty(dir_path):
-            self.output.prompt("Directory is not empty. Run 'iotedgedev azure' or clean the directory.")
+            self.output.prompt("Directory is not empty. Run 'iotedgedev iothub' or clean the directory.")
```
jongio: See comment above for why azure.

iotedgedev/solution.py
```diff
@@ -33,7 +33,7 @@ def create(self, name, module, template):
 
         os.rename(os.path.join(name, ".env.tmp"), os.path.join(name, ".env"))
 
-        mod_cmd = "iotedgedev addmodule {0} --template {1}".format(module, template)
+        mod_cmd = "iotedgedev solution add {0} --template {1}".format(module, template)
```
jongio: Should this be addmodule or just add? This might be misinterpreted as "add a solution"

iotedgedev/azurecli.py
LazarusX: There is a `monitor` subcommand under the `iothub` category which is used to monitor D2C messages. I think `iotedgedev azure monitor` sounds a little bit weird. If we include container registry feature in the future, we can put it under the `docker` category or add a new `acr` category (which follows Azure CLI's style).

iotedgedev/solution.py
LazarusX: This is a tradeoff between convenience and expressiveness. Adding modules should be a frequently used command (more than the create solution command), so making it shorter can save keystrokes. On the other hand, given the existence of the `create` command and the help messages, I suppose the misinterpretation to be unlikely.

iotedgedev/azurecli.py
jongio: Okay

jongio:
Please update all associated docs.  Readme.  Wiki.

LazarusX:
@jongio Sure. Issue https://github.com/Azure/iotedgedev/issues/201 is updated with the items need to be documented.

Issue #213, opened 2018-07-31
Need to throw error when deployment manifest is break


michaeljqzq:
For example, the following manifest
```json
{
  "moduleContent": {
    "$edgeAgent": {
    },
    "$edgeHub": {
      "properties.desired": {
        "schemaVersion": "1.0",
        "routes": {
          "sensorToFilter1": "FROM /messages/modules/temp-sensor-module/outputs/temperatureOutput INTO BrokeredEndpoint(\"/modules/filtermodule/inputs/input1\")",
          "filter2": "FROM /messages/modules/filtermodule/outputs/output1 INTO BrokeredEndpoint(\"/modules/filtermodule2/inputs/input1\")",
          "filter3": "FROM /messages/modules/filtermodule2/outputs/output1 INTO BrokeredEndpoint(\"/modules/filtermodule3/inputs/input1\")",
          "filterfunction": "FROM /messages/modules/filtermodule3/outputs/output1 INTO BrokeredEndpoint(\"/modules/functionmodule/inputs/input1\")",
          "filterToIoTHub": "FROM /messages/modules/functionmodule/outputs/output1 INTO $upstream"
        },
        "storeAndForwardConfiguration": {
          "timeToLiveSecs": 7200
        }
      }
    },
    "filtermodule": {
      "properties.desired": {
        "schemaVersion": "1.0",
        "TemperatureThreshold": 21
      }
    }
  }
}
```

Currently, run "iotedgedev push" will return
```
Running: iotedgedev push

=======================================
======== ENVIRONMENT VARIABLES ========
=======================================

Environment Variables loaded from: .env (C:\Users\v-zhq\AppData\Local\Temp\hudson186348797278582691test\workspace\test0\.env)

=========================================
======== PROCESSING CONFIG FILES ========
=========================================

Expanding 'deployment.template.json' to 'config\deployment.json'


=================================================
======== INITIALIZING CONTAINER REGISTRY ========
=================================================

REGISTRY: michaeljqzq


==================================
======== BUILDING MODULES ========
==================================


=========================================
======== PROCESSING CONFIG FILES ========
=========================================

Expanding 'deployment.template.json' to 'config\deployment.json'


Finished: SUCCESS
```

Issue #212, opened 2018-07-30
Support new schema of deployment.json with backward compatibility


LazarusX:
`moduleContent` in deployment.json is now `modulesContent`: https://docs.microsoft.com/en-us/azure/iot-edge/how-to-deploy-modules-cli

PR #222, opened 2018-08-02
Support new schema of deployment.json with backward compatibility

LazarusX:
This PR implements https://github.com/Azure/iotedgedev/issues/212 and fixes https://github.com/Azure/iotedgedev/issues/213.

### DIFF
--- a/iotedgedev/deploymentmanifest.py
+++ b/iotedgedev/deploymentmanifest.py
@@ -43,7 +43,11 @@ class DeploymentManifest:
             }
         }"""
 
-        self.utility.nested_set(self.json, ["moduleContent", "$edgeAgent", "properties.desired", "modules", module_name], json.loads(new_module))
+        try:
+            self.utility.nested_set(self.get_module_content(), ["$edgeAgent", "properties.desired", "modules", module_name], json.loads(new_module))
+        except KeyError as err:
+            self.output.error("Missing key {0} in file {1}".format(err, self.path))
+            sys.exit(1)
 
         self.add_default_route(module_name)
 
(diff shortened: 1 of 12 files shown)


iotedgedev/deploymentmanifest.py
```diff
@@ -52,24 +56,58 @@ def add_default_route(self, module_name):
         new_route_name = "{0}ToIoTHub".format(module_name)
         new_route = "FROM /messages/modules/{0}/outputs/* INTO $upstream".format(module_name)
 
-        self.utility.nested_set(self.json, ["moduleContent", "$edgeHub", "properties.desired", "routes", new_route_name], new_route)
+        try:
+            self.utility.nested_set(self.get_module_content(), ["$edgeHub", "properties.desired", "routes", new_route_name], new_route)
```
jongio: Seems like we can further abstract "get desired property" which takes module and name.

tests/assets/deployment.template_2.json
```diff
@@ -29,7 +29,7 @@
           }
         },
         "modules": {
-          "temp-sensor-module": {
+          "tempSensor": {
```
jongio: I used all lower because docker is all lower and I use that as the docker name.  Thoughts?

LazarusX: @jongio The documents all reference to the temp sensor module as "tempSensor", for example https://docs.microsoft.com/en-us/azure/iot-edge/quickstart-linux#deploy-a-module and https://docs.microsoft.com/en-us/azure/iot-edge/tutorial-csharp-module#build-your-iot-edge-solution. For container names, Docker supports upper case. The edgeAgent and edgeHub also follows this style.

jongio: @LazarusX  - This is what I get when I run build with mixed case

```
(venv) C:\temp>docker build . -t testFile
invalid argument "testFile" for "-t, --tag" flag: invalid reference format: repository name must be lowercase
See 'docker build --help'.
```

iotedgedev/deploymentmanifest.py
LazarusX: @jongio Added. Thanks.

tests/assets/deployment.template_2.json
jongio: Okay.

jongio:
It would be good to add tests for the new get_desired_property method.

LazarusX:
@jongio Tests added. Thanks.

jongio:
@LazarusX  - you ready for merge on this?

LazarusX:
@jongio Yes, you can merge it if it looks good to you.

PR #230, opened 2018-08-03
Add envvars.clean to work around the py2 unicode in envvars issue.

jongio:


acumino:
Please share the solution.

PR #255, opened 2018-08-15
Support build options

w12379564:
Parse the docker cli build command options to python sdk dict.

### DIFF
--- a/iotedgedev/buildoptions.py
+++ b/iotedgedev/buildoptions.py
@@ -68,6 +68,9 @@ def split_kv(kvpair):
         return kvpair, ''
 
 
+# Mapping docker build cli command to docker python sdk.
+# Format: cli_key : [sdk_key, parse_func]
+# The commented items are cli build options not supported in python sdk.
 cli_sdk_mapping = {
     '--add-host': ['extra_hosts', functools.partial(parse_to_dict, split_host)],
     '--build-arg': ['buildargs', functools.partial(parse_to_dict, split_arg)],
@@ -105,6 +108,8 @@ cli_sdk_mapping = {
     '--target': ['target', parse_val],
     # '--ulimit' : ['sdk_key', parse_func]
 }
+
+# Class BuildOptions parse docker cli build commands to python sdk dict.
 
 
 class BuildOptions(object):
--- a/iotedgedev/buildoptions.py
+++ b/iotedgedev/buildoptions.py
@@ -1,0 +1,110 @@
+import functools
+import re
+
+
+def parse_to_dict(split_func, sdk_options, sdk_key, build_option_val):
+    if sdk_key not in sdk_options:
+        sdk_options[sdk_key] = {}
+    k, v = split_func(build_option_val)
+    sdk_options[sdk_key][k] = v
+
+
+def parse_to_list(sdk_options, sdk_key, build_option_val):
+    if sdk_key not in sdk_options:
+        sdk_options[sdk_key] = []
+    sdk_options[sdk_key].append(build_option_val)
+
+
+def parse_val(sdk_options, sdk_key, build_option_val):
+    sdk_options[sdk_key] = build_option_val
+
+
+def parse_container_limits(sdk_options, sdk_key, build_option_val):
+    if 'container_limits' not in sdk_options:
+        sdk_options['container_limits'] = {}
+    sdk_options['container_limits'][sdk_key] = build_option_val
+
+
+def parse_flag(sdk_options, sdk_key, build_option_val):
+    sdk_options[sdk_key] = True
+
+
+def split_build_option(build_option_str):
+    parsed_option = re.compile(r"\s+").split(build_option_str)
+    if len(parsed_option) == 1:
+        parsed_option = parsed_option[0].split('=', 1)
+    if len(parsed_option) == 1:
+        return parsed_option[0], None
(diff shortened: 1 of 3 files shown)


LazarusX:
This PR closes #255.

Hi, @jongio, our intern @w12379564 helped add build options support for iotedgedev. This PR convert the build options in module.json, which is in Docker CLI syntax, to the format which docker-py accepts. Please help review. Thanks!

iotedgedev/buildoptions.py
```diff
@@ -0,0 +1,144 @@
...
+    if '=' in kvpair:
+        return kvpair.split('=', 1)
+    else:
+        return kvpair, ''
+
+
+# Mapping docker build cli command to docker python sdk.
+# Format: cli_key : [sdk_key, parse_func]
+# The commented items are cli build options not supported in python sdk.
+cli_sdk_mapping = {
```
digimaun: assuming all this mapping from cli to sdk element is valid

iotedgedev/buildoptions.py
```diff
@@ -0,0 +1,144 @@
...
+    '--quiet': ['quiet', parse_flag],
+    '-q': ['quiet', parse_flag],
+    '--rm': ['rm', parse_flag],
+    # '--security-opt' : ['sdk_key', parse_func],
+    '--shm-size': ['shmsize', parse_val],
+    '--squash': ['squash', parse_flag],
+    # '--stream' : ['sdk_key', parse_func],
+    '--tag': ['tag', parse_val],
+    '-t': ['tag', parse_val],
+    '--target': ['target', parse_val],
```
digimaun: linter may catch comma on last item (just pointing out)

iotedgedev/buildoptions.py
```diff
@@ -0,0 +1,144 @@
...
+
+# Class BuildOptions parse docker cli build commands to python sdk dict.
+
+
+class BuildOptions(object):
+    def __init__(self, build_options):
+        self.build_options = build_options
+        self.sdk_options = {}
+
+    def filter_build_options(self, build_options):
```
digimaun: if filter_build_options is classified as 'private' then prefix the function with an underscore _

iotedgedev/buildoptions.py
```diff
@@ -0,0 +1,144 @@
...
+        """Remove build options which will be ignored"""
+        if build_options is None:
+            return None
+
+        filtered_build_options = []
+        for build_option in build_options:
+            build_option = build_option.strip()
+            parsed_option = re.compile(r"\s+").split(build_option)
+            if parsed_option and parsed_option[0] not in ["--rm", "--tag", "-t", "--file", "-f"]:
```
digimaun: extract the filter set to something more reusable like a class variable

digimaun: consider warning the user if some options passed in are ignored

iotedgedev/buildoptions.py
```diff
@@ -0,0 +1,144 @@
...
+                filtered_build_options.append(build_option)
+
+        return filtered_build_options
+
+    def parse_build_options(self):
+        """Parse build options to python SDK"""
+        filtered_build_options = self.filter_build_options(self.build_options)
```
digimaun: Here there is passing of an instance variable to an instance method. Because the class requires and sets build_options in the constructor and parse_build_options doesn't accept an argument, it makes sense for filter_build_options to also not require build_options as an argument and assume operation on the instance variable.

tests/test_buildoptions.py
```diff
@@ -0,0 +1,118 @@
...
+def test_filter_build_options():
+    build_options = [
+        "--rm",
+        "-f test",
+        "--file test",
+        "-t image",
+        "--tag image"
+    ]
+    build_options_parser = BuildOptions(build_options)
+    assert not build_options_parser.filter_build_options(build_options)
```
digimaun: See comment in iotedgedev/buildoptions.py line 136, the same information (build_options) is being passed in twice when it could be once

iotedgedev/modules.py
```diff
@@ -131,7 +133,7 @@ def build_push(self, no_build=False, no_push=False):
                         dockerfile = dockerfile.replace("\\", "/")
                         dockerfile_relative = dockerfile_relative.replace("\\", "/")
 
-                    build_result = self.dock.docker_client.images.build(tag=tag, path=context_path, dockerfile=dockerfile_relative)
+                    build_result = self.dock.docker_client.images.build(tag=tag, path=context_path, dockerfile=dockerfile_relative, **sdk_options)
```
digimaun: nice

tests/test_buildoptions.py
```diff
@@ -0,0 +1,118 @@
+import pytest
```
digimaun: consider adding a test for filtered and valid build options in the same parse call

w12379564:
Hi @digimaun , thanks for your review! I have made some updates according to your comments:
* Rename the filter_build_options with prefix _
* Extract the filter set and provide warning if build options are ignored
* _filter_build_options now doesn't require build_options argument
* Refine the filter logic
* Add a test for filtered and valid build options
* Resolve conflicts

jongio:
Thanks @w12379564 - @digimaun - will merge after you sign off.

digimaun:
I like the changes @w12379564 but please resolve the new conflict.

PR #263, opened 2018-08-17
Fixed minor container problems

yorek:
updated scripts to reflex new docker folder structure
merged with latest updates from upstream repo
tox tests in container still failing due to new (?) package "commentjson" used somewhere?


docker/tool/linux/Dockerfile
```diff
@@ -2,4 +2,5 @@ FROM iotedgedev-linux-base
 ARG IOTEDGEDEV_VERSION
 COPY iotedgedev-$IOTEDGEDEV_VERSION-py2.py3-none-any.whl dist/iotedgedev-$IOTEDGEDEV_VERSION-py2.py3-none-any.whl
 RUN pip install dist/iotedgedev-$IOTEDGEDEV_VERSION-py2.py3-none-any.whl
-
+RUN export LC_ALL=C.UTF-8 && \
```
jongio: Can't this be in linux-base image?

docker/tool/linux/run.ps1
```diff
@@ -1,5 +1,5 @@
 # Get IoTEdgeDev source folder
-$source_folder = Get-Location | Split-Path | Split-Path 
+$source_folder = Get-Location | Split-Path | Split-Path | Split-Path
 
 # Run Docker Container
 docker run -it -v //var//run//docker.sock://var//run//docker.sock -v ${source_folder}:/home/iotedge/tool microsoft/iotedgedev:latest-linux
```
jongio: When I run the container I map to /home/iotedge root because that is where I want to store files by default.

1. keep the tool directory mapping
2. add a second mapping that is the working directory...the default directory that the user is dropped into when they run the container.

docker/tool/linux/run.ps1
```diff
@@ -1,5 +1,5 @@
 # Get IoTEdgeDev source folder
-$source_folder = Get-Location | Split-Path | Split-Path 
+$source_folder = Get-Location | Split-Path | Split-Path | Split-Path
```
jongio: Can you also get the run.sh?

LazarusX:
"commentjson" is a new dependency added by me. Could you try if removing `.tox` folder can fix the tox failure?

docker/tool/linux/run.ps1
jongio: I realized there are two personas that was diff mapped drives.

1. Tool developer who needs to run tox in container
2. End user who wants to map /home/iotedge to a local folder.

Can you create a diff run file for each?  

One that maps current folder to /home/iotedge for #1 above
One that maps current folder to /home/iotedge/tool for #2 above.

jongio:
Tox passes successfully.  Merging.  We can discuss run-dev/run scenario after this weeks release.

PR #274, opened 2018-08-21
Add option to setup before starting simulator

LazarusX:


jongio:
Build failed on py3, restarted it.

PR #325, opened 2018-09-25
Support adding C modules

### LINKED ISSUE #102: Support C Modules

LazarusX:
Close #102

### DIFF
--- a/iotedgedev/modules.py
+++ b/iotedgedev/modules.py
@@ -1,5 +1,6 @@
 import os
 import re
+import shutil
 import sys
 from zipfile import ZipFile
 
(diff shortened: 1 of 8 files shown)


iotedgedev/modules.py
```diff
@@ -36,10 +41,24 @@ def add(self, name, template):
...
 
-        deployment_manifest = DeploymentManifest(self.envvars, self.output, self.utility, self.envvars.DEPLOYMENT_CONFIG_TEMPLATE_FILE, True)
-
         repo = "{0}/{1}".format("${CONTAINER_REGISTRY_SERVER}", name.lower())
-        if template == "csharp":
+        if template == "c":
+            github_prefix = "https://github.com/Azure"
+            git_repo = "azure-iot-edge-c-module"
+            branch = "master"
+            url = "{0}/{1}/archive/{2}.zip".format(github_prefix, git_repo, branch)
```
jongio: Are we doing this because Nuget was removed from the C SDK repo?

iotedgedev/modules.py
```diff
@@ -36,10 +41,24 @@ def add(self, name, template):
...
+            response = urlopen(url)
```
jongio: Thoughts on caching this zip locally?  Have you tested the download time of this zip file?  Does it get all recursive projects? Do we need those projects?

LazarusX: The size of the zip file is only 8KB, so we'd better always fetch the latest instead of caching them. The dependencies are installed via `apt-get` in the [Dockerfile](https://github.com/Azure/azure-iot-edge-c-module/blob/master/Dockerfile.amd64#L6)

iotedgedev/modules.py
formulahendry: No related to the distribution of SDK. For Node scaffolding, we use Yeoman. For python scaffolding, we use Cookiecutter. For C, there is no appropriate scaffolding tool, so we use github to host the C template fot IoT Edge Module.

jongio: Sorry, my comments were due to me thinking this was the actual IoT C SDK.

PR #328, opened 2018-09-30
Collect IoT Hub hostname hash and suffix when deploying

LazarusX:
This is a BI requirement.

@chyuanmsft, please help take a look.

### DIFF
--- a/iotedgedev/connectionstring.py
+++ b/iotedgedev/connectionstring.py
@@ -3,11 +3,11 @@ from .utility import Utility
 
 class ConnectionString:
     def __init__(self, value):
-        self.ConnectionString = value
+        self.connection_string = value
         self.data = dict()
 
-        if self.ConnectionString:
-            parts = self.ConnectionString.split(';')
+        if self.connection_string:
+            parts = self.connection_string.split(';')
             if len(parts) > 0:
                 for part in parts:
                     subpart = part.split('=', 1)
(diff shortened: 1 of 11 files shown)


iotedgedev/utility.py
```diff
@@ -188,3 +189,22 @@ def nested_set(self, dic, keys, value):
...
+    def hash_connection_str_hostname(connection_str_hostname):
+        """Hash connection string hostname to count distint IoT Hub number"""
+
+        # get hostname suffix (e.g., azure-devices.net) to distinguish national clouds
+        if "." in connection_str_hostname:
+            hostname_suffix = connection_str_hostname[connection_str_hostname.index(".")+1:]
+        else:
+            hostname_suffix = ""
+
+        return (Utility.get_sha256_hash(connection_str_hostname), hostname_suffix)
```
chyuanmsft: Same with another PR, we expect to record empty string when host name is empty.

LazarusX: Updated. Please help check again.

iotedgedev/utility.py
```diff
@@ -188,3 +189,24 @@ def nested_set(self, dic, keys, value):
             current = current.get(key)
 
         current[keys[-1]] = value
+
+    @staticmethod
+    @suppress_all_exceptions()
+    def hash_connection_str_hostname(connection_str_hostname):
```
jongio: Seems like this could be in the ConnectionString class.

LazarusX: Moved. Thanks.

iotedgedev/connectionstring.py
```diff
@@ -20,6 +24,22 @@ def __init__(self, value):
     def __getitem__(self, key):
         return self.data[key]
 
+    @suppress_all_exceptions()
+    def hash_hostname(self):
```
jongio: I was thinking this could be calculated when the connection string was initially parsed.

so you can just access it with `connection_string.HostNameHashed`

LazarusX: Updated. Should look better now.

iotedgedev/connectionstring.py
```diff
@@ -13,8 +16,16 @@ def __init__(self, value):
 
             if self.data:
                 self.HostName = self["hostname"]
-                if self.HostName:
+                if self.HostName and "." in self.HostName:
```
jongio: Alternatively you could expand "Host"  to a class, so you can have: Host.Name, Host.NameHashed, Host.Suffix

LazarusX: Thanks. Updated.

PR #353, opened 2018-11-01
Sleep 3s after running local registry to ensure it can accept push requests

LazarusX:
Look like it take a few seconds for the local registry to start accepting push requests after it starts running.


LazarusX:
Close it for now. Will investigate a more graceful method.

PR #357, opened 2018-11-09
Support scaffolding and parsing createOptions as a dict

### LINKED ISSUE #354: Support chunking long createOptions
Ref: https://github.com/Azure/iotedge/pull/396

### LINKED ISSUE #356: Support createOptions in JSON object format

LazarusX:
Close #356 and close #354

### DIFF
--- a/iotedgedev/deploymentmanifest.py
+++ b/iotedgedev/deploymentmanifest.py
@@ -5,12 +5,18 @@ and deployment manifest template (deployment.template.json)
 
 import json
 import os
+import re
 import shutil
+
+import six
 
 from .compat import PY2
 
 if PY2:
     from .compat import FileNotFoundError
+
+TWIN_VALUE_MAX_SIZE = 512
+TWIN_VALUE_MAX_CHUNKS = 8
 
 
 class DeploymentManifest:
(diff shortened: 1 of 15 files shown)


iotedgedev/modules.py
```diff
@@ -155,7 +155,7 @@ def build_push(self, no_build=False, no_push=False):
             key = (module_name, platform)
             if key in image_tag_map:
                 tag = image_tag_map.get(key)
-                replacements["${{MODULES.{0}.{1}}}".format(module_name, platform)] = tag
+                replacements[module_name] = tag
```
adashen: one question here. If the module didn't use this schema, will it also be replaced?

tests/assets/deployment.template_3.json
```diff
@@ -37,7 +57,70 @@
             "restartPolicy": "always",
             "settings": {
               "image": "mcr.microsoft.com/azureiotedge-simulated-temperature-sensor:1.0",
-              "createOptions": "{}"
+              "createOptions": {
```
adashen: We could still keep some test cases with string value. We should support both types in template

iotedgedev/deploymentmanifest.py
```diff
@@ -100,14 +111,42 @@ def get_modules_to_process(self):
...
+    def convert_create_options(self):
+        modules = self.get_all_modules()
+        for module_name, module_info in modules.items():
+            if "settings" in module_info and "createOptions" in module_info["settings"]:
+                create_options = module_info["settings"]["createOptions"]
+                if not isinstance(create_options, six.string_types):
+                    # Stringify and minify the createOptions from dict format
+                    create_options = json.dumps(create_options, separators=(',', ':'))
+
+                options = [m for m in re.finditer("(.|[\r\n]){{1,{0}}}".format(TWIN_VALUE_MAX_SIZE), create_options)]
```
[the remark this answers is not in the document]
LazarusX: The format string use double {} to escape {}.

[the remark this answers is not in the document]
adashen: got it

iotedgedev/modules.py
LazarusX: No, see https://github.com/Azure/iotedgedev/blob/master/iotedgedev/deploymentmanifest.py#L84-L100

adashen: Can we use regex to get the to be process modules?

PR #358, opened 2018-11-21
Support multi-platform deployment manifest template

LazarusX:


LazarusX:
@michaeljzq, FYI

iotedgedev/cli.py
```diff
@@ -122,10 +121,17 @@ def e2e(ctx):
...
+@click.option("--file",
+              "-f",
+              "template_file",
+              default=envvars.DEPLOYMENT_CONFIG_TEMPLATE_FILE,
+              show_default=True,
+              required=False,
+              help="Specify the deployment manifest template file")
 @with_telemetry
-def add(name, template, group_id):
+def add(name, template, group_id, template_file):
```
adashen: will the module be added to debug template also?

iotedgedev/cli.py
```diff
@@ -147,11 +153,24 @@ def add(name, template, group_id):
...
+@click.option("--file",
+              "-f",
+              "template_file",
+              default=envvars.DEPLOYMENT_CONFIG_TEMPLATE_FILE,
+              show_default=True,
+              required=False,
+              help="Specify the deployment manifest template file")
+@click.option("--platform",
+              "-P",
+              default=envvars.DEFAULT_PLATFORM,
```
adashen: we don't have hard code default value? It seems not compatible if user use the new version tool to build solution created previously

LazarusX: The default value for `DEFAULT_PLATFORM` is amd64 (loaded in https://github.com/Azure/iotedgedev/pull/358/files#diff-1fc02f4fdd45b74526b9cf54754645ceR127). It will be ignored if the user is building a legacy solution.

iotedgedev/cli.py
LazarusX: Yes, in https://github.com/Azure/iotedgedev/pull/358/files#diff-81a619e7b82430890d648131b2db0338R111

iotedgedev/cli.py
```diff
@@ -273,14 +338,14 @@ def setup_simulator(gateway_host):
               show_default=True,
               help="Port of the service for sending message.")
 @with_telemetry
-def start_simulator(setup, solution, build, verbose, inputs, port):
+def start_simulator(setup, solution, build, template_file, platform, verbose, inputs, port):
```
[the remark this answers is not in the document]
LazarusX: Only when the `build` flag is set.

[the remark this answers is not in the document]
adashen: the template_file is only valid when is build ?

iotedgedev/cli.py
jongio: Or make it an override and default to the env var setting.

.env.tmp
```diff
@@ -54,11 +54,13 @@ CONTAINER_TAG=""
 #
 
 CONFIG_OUTPUT_DIR="config"
-DEPLOYMENT_CONFIG_FILE="deployment.json"
 DEPLOYMENT_CONFIG_TEMPLATE_FILE="deployment.template.json"
-LOGS_PATH="logs"
+DEPLOYMENT_CONFIG_DEBUG_TEMPLATE_FILE="deployment.debug.template.json"
```
jongio: I think this file should be named `deployment.template.debug.json`.

iotedgedev/deploymentmanifest.py
```diff
@@ -21,41 +22,42 @@
...
 class DeploymentManifest:
     def __init__(self, envvars, output, utility, path, is_template):
+        self.envvars = envvars
         self.utility = utility
         self.output = output
         try:
             self.path = path
             self.is_template = is_template
-            self.json = json.loads(self.utility.get_file_contents(path, expandvars=True))
+            self.json = json.loads(Utility.get_file_contents(path, expandvars=True))
```
jongio: Why the change to Utility and not use self.utility?

iotedgedev/dockercls.py
```diff
@@ -127,7 +132,7 @@ def setup_registry_in_config(self, image_names):
 
         # Replace mcr.microsoft.com/ with ${CONTAINER_REGISTRY_SERVER}
         for config_file in self.utility.get_config_files():
-            config_file_contents = self.utility.get_file_contents(config_file)
+            config_file_contents = Utility.get_file_contents(config_file)
```
jongio: Same Utility comment.

.env.tmp
LazarusX: We would like to allow users pass deployment manifest templates with any name, but we also want to set some rules so that we can determine what the generated deployment manifest would be named. The current rule is that if the template is XXX.template.json, the generated one will be named config/XXX.json. So when deployment.debug.template.json is built, the generated file would be config/deployment.debug.json` (the naming is implemented at https://github.com/Azure/iotedgedev/pull/358/files#diff-4a4d85e50c4c4f0af9d1ca17c4252649R166)

iotedgedev/cli.py
LazarusX: @jongio It already defaults to env var setting (https://github.com/Azure/iotedgedev/pull/358/files#diff-f7f95b5a237aedfa9c80a3ec57d525e7R127). I am working on updating it to default to `envvars.DEPLOYMENT_CONFIG_FILE_PATH`, so that users can specify a built deployment manifest.

PR #372, opened 2019-01-25
Fix #371

### LINKED ISSUE #371: iotedgedev genconfig applies lowercase on environment variables
### Problem
`iotedgedev genconfig` applies lowercase on environment variables when generating the deployment manifest from the template.

### Description
Our team is using the Git commit tag name as docker image label for the IoT edge module. This label is set as an environment variable in the `deployment.template.json`

When using the 'Generate IoT Edge Deployment Manifest' from the context menu in VS Code, the deployment manifest is generated with the docker image label in the right case (the tag `1.2.3-RC1` is used as-is for the docker image label).

When using the `iotedgedev genconfig` command, our Git commit tag becomes lowercase (`1.2.3-RC1` becomes `1.2.3-rc1`), which results in failed deployments, as the edgeAgent cannot fetch the images from the container registry.

### Suggested solution
Environment variables in uppercase should be allowed

Thank you for your help.

LazarusX:


iotedgedev/modules.py
```diff
@@ -145,7 +145,7 @@ def build_push(self, template_file, default_platform, no_build=False, no_push=Fa
                     # get the Dockerfile from module.json
                     dockerfile = module.get_dockerfile_by_platform(platform)
                     container_tag = "" if self.envvars.CONTAINER_TAG == "" else "-" + self.envvars.CONTAINER_TAG
-                    tag = "{0}:{1}{2}-{3}".format(module.repository, module.tag_version, container_tag, platform).lower()
+                    tag = "{0}:{1}{2}-{3}".format(module.repository, module.tag_version, container_tag, platform)
```
jongio: FYI for future reviewers...this comment can be closed.

Image name needs to be lower case.  module.repository = {server}/{name}, so you need to call .lower on that, but not on the tag.  I have confirmed that name.lower is called earlier in the code.

vsts_ci/win32/continuous-build-win32.yml
```diff
@@ -3,33 +3,38 @@ steps:
...
+  - task: DotNetCoreInstaller@0
+    inputs:
+      packageType: 'sdk'
+      version: '2.2.103'
+    displayName: "Install .NET Core SDK"
+
   - powershell: |
-      choco install jdk11
-      $env:JAVA_HOME = "C:\Program Files\Java\jdk-11.0.1"
+      choco install jdk8 --version 8.0.201
```
LazarusX: Switch to JDK8 because it's already specified as a dependency of maven. Installing JDK11 will cost more time and space.

LazarusX:
Currently CI build will fail on Windows agents only because Windows agents are still based on 1803 while the scaffolded Dockerfiles are based on 1809 (Azure DevOps' 1809 support is coming by end of Feb).

iotedgedev/modules.py
```diff
@@ -171,57 +172,58 @@ def build_push(self, template_file, default_platform, no_build=False, no_push=Fa
...
-                        self.output.info("Could not find registry credentials with name {0} in environment variable. Pushing anonymously.".format(tag.split('/')[0].lower()))
-                        response = docker.docker_client.images.push(repository=tag, stream=True)
-                    else:
-                        response = docker.docker_client.images.push(repository=tag, stream=True, auth_config={
-                            "username": self.envvars.CONTAINER_REGISTRY_MAP[registry_key].username,
-                            "password": self.envvars.CONTAINER_REGISTRY_MAP[registry_key].password})
-                    docker.process_api_response(response)
-            self.output.footer("BUILD COMPLETE", suppress=no_build)
-            self.output.footer("PUSH COMPLETE", suppress=no_push)
+        if not no_build or not no_push:
```
adashen: so it will build even if user specify no_build but push?

LazarusX: No. There are extra check in Line 180 and Line 206. Adding check in Line 175 just to avoid creating `Docker` object on Line 176 when users only want to genconfig.

setup.py
```diff
@@ -32,9 +32,10 @@ def run(self):
     'Click>=6.0',
     'docker>=3.5',
     'python-dotenv',
-    'requests',
+    'requests<2.21',
```
adashen: what is the issue if use higher version?

setup.py
```diff
@@ -32,9 +32,10 @@ def run(self):
...
     'fstrings',
     'msrestazure~=0.4.32',
+    'azure-cli-core<2.0.55',
```
adashen: same as above

setup.py
adashen: And can we make sure what is the lowest version supported?

setup.py
LazarusX: `azure-cli-core` depends on a beta version of PyYAML starting version 2.0.55, which will conflict with docker-compose. See https://github.com/Azure/iotedgedev/issues/373

LazarusX: https://github.com/Azure/azure-cli/blame/dev/src/azure-cli-core/setup.py#L71

setup.py
LazarusX: Similar conflict with docker-compose https://github.com/docker/compose/blob/master/setup.py#L36

LazarusX: Lower bound specified.

setup.py
LazarusX: Lower bound specified.

PR #382, opened 2019-03-21
Fix host.docker.internal not available in linux docker issue

blackchoey:
host.docker.internal is not available in docker for linux, use the actual ip address when host.docker.internal is unavailable


blackchoey:
#380

Issue #383, opened 2019-04-11
incompatible applicationinsights 0.11.8 dependency 


cicorias:
- iotedgedev Version:  1.2
- Python Version: 3.6.8
- Pip Version: 18.1
- Development machine OS Version: Debin/Ubuntu - WSL Pengwin
- IoT Edge device OS Version: Ubuntu

Steps to Reproduce:

1.  Using a pyhton installation of 3.6.8 - either system level or via venv
2. run `pip install -U iotedgedev

Output is

```
azure-cli-telemetry 1.0.2 has requirement applicationinsights<0.11.8,>=0.11.1, but you'll have applicationinsights 0.11.8 which is incompatible.

```

![image](https://user-images.githubusercontent.com/1080406/55977240-4bcd4080-5c5c-11e9-958d-d18d64a7d5ab.png)

Azure cli 

![image](https://user-images.githubusercontent.com/1080406/55977266-63a4c480-5c5c-11e9-963e-c1d80e3b7437.png)

```
azure-cli                         2.0.62

acr                                2.2.4
acs                               2.3.21
advisor                            2.0.0
ams                                0.4.4
appservice                        0.2.17
backup                             1.2.3
batch                              4.0.0
batchai                            0.4.8
billing                            0.2.1
botservice                        0.1.10
cdn                                0.2.2
cloud                              2.1.1
cognitiveservices                  0.2.5
command-modules-nspkg               2.0.2
configure                         2.0.21
consumption                        0.4.2
container                         0.3.16
core                              2.0.62
cosmosdb                           0.2.9
dla                                0.2.5
dls                                0.1.8
dms                                0.1.3
eventgrid                          0.2.2
eventhubs                          0.3.4
extension                          0.2.5
feedback                           2.2.0
find                               0.3.2
hdinsight                          0.3.3
interactive                        0.4.3
iot                                0.3.7
iotcentral                         0.1.6
keyvault                          2.2.14
kusto                              0.2.2
lab                                0.1.6
maps                               0.3.4
monitor                           0.2.12
network                            2.3.6
nspkg                              3.0.3
policyinsights                     0.1.2
privatedns                         1.0.0
profile                            2.1.5
rdbms                              0.3.9
redis                              0.4.2
relay                              0.1.4
reservations                       0.4.2
resource                          2.1.13
role                               2.5.0
search                             0.1.1
security                           0.1.1
servicebus                         0.3.4
servicefabric                     0.1.16
signalr                            1.0.0
[...]

Issue #373, opened 2019-01-29
Fail to run after new installation


LazarusX:
`pkg_resources.ContextualVersionConflict: (PyYAML 3.13 (c:\venv\lib\site-packages), Requirement.parse('pyyaml>=4.2b1'), {'azure-cli-core'})`

LazarusX:
New error appeared on CI after upgrading to `pyyaml>=4,2b1` in https://github.com/Azure/iotedgedev/pull/372/commits/d6436e102ceccc0e2d54d602b80e5313b5481745:
```
pkg_resources.ContextualVersionConflict: (PyYAML 4.2b4 (/mnt/a/1/s/.tox/py37/lib/python3.7/site-packages), Requirement.parse('PyYAML<4,>=3.10'), {'docker-compose'})
```

jongio:
Aren’t we using explicit versions for all dependencies? We shouldn’t run into these types of issues if so, right?

LazarusX:
No, we aren't now.

I was considering locking the all dependency versions (https://github.com/Azure/iotedgedev/compare/ray-deps?expand=1) after reading [this piece](https://blog.miguelgrinberg.com/post/the-package-dependency-blues), but realized that there will be more version conflicts if the user installs iotedgedev in the global environment. I prefer specifying a max version of azure-cli-core (https://github.com/Azure/iotedgedev/pull/372/commits/4aed1c3854f5e1874e9df2388f18a84126c2c58c)

LazarusX:
Specifying version ranges for all dependencies like docker-compose does (https://github.com/docker/compose/blob/master/setup.py#L32-L43) will definitely help, but it's surprising that azure-cli-core upgraded PyYAML to a new major beta version in its patch release.

LazarusX:
Should be fixed by https://github.com/Azure/iotedgedev/pull/372

PR #384, opened 2019-04-18
Support iotedgehubdev new setup parameter

blackchoey:
1. Support iotedgehubdev new setup parameter `-i`
2. Update applicationinsights dependency to avoid breaking installation (#383)
3. Update azure-cli-core dependency to avoid breaking deploy function (error loading command module cloud)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ requirements = [
     'azure-cli-configure',
     'azure-cli-resource',
     'azure-cli-cloud',
-    'iotedgehubdev',
+    'iotedgehubdev >= 0.8.0',
     'six',
     'applicationinsights',
     'commentjson',
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,7 @@ requirements = [
     'requests >= 2.20.0, < 2.21',
     'fstrings',
     'msrestazure~=0.4.32',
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -35,16 +35,16 @@ def run(self):
     'requests >= 2.20.0, < 2.21',
     'fstrings',
     'msrestazure~=0.4.32',
-    'azure-cli-core >= 2.0.0, <2.0.55',
+    'azure-cli-core >= 2.0.0, <=2.0.62',
```
adashen: depends on a beta version of PyYAML starting version 2.0.55, which will conflict with docker-compose. See #373 
Is this  issue avoided?

blackchoey: Yes
1. The latest azure-cli-core doesn't specify version for pyyaml now [https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/setup.py#L70](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/setup.py#L70)
2. docker-compose also bumped pyyaml version range to >= 3.10, < 4.3 [https://github.com/docker/compose/blob/master/setup.py#L35](https://github.com/docker/compose/blob/master/setup.py#L35)

Issue #387, opened 2019-04-27
KeyError: 'cs' in azurecli.py


jporcenaluk:
There is an issue (Windows 10, python 3.7, Azure CLI) where when I run the setup, I am greeted with the following output:

```
Traceback (most recent call last):
  File "c:\users\jporc\appdata\local\programs\python\python37-32\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "c:\users\jporc\appdata\local\programs\python\python37-32\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Users\jporc\AppData\Local\Programs\Python\Python37-32\Scripts\iotedgedev.exe\__main__.py", line 9, in <module>
  File "c:\users\jporc\appdata\local\programs\python\python37-32\lib\site-packages\click\core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "c:\users\jporc\appdata\local\programs\python\python37-32\lib\site-packages\click\core.py", line 717, in main
    rv = self.invoke(ctx)
  File "c:\users\jporc\appdata\local\programs\python\python37-32\lib\site-packages\click\core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "c:\users\jporc\appdata\local\programs\python\python37-32\lib\site-packages\click\core.py", line 1135, in invoke
    sub_ctx = cmd.make_context(cmd_name, args, parent=ctx)
  File "c:\users\jporc\appdata\local\programs\python\python37-32\lib\site-packages\click\core.py", line 641, in make_context
    self.parse_args(ctx, args)
  File "c:\users\jporc\appdata\local\programs\python\python37-32\lib\site-packages\click\core.py", line 940, in parse_args
    value, args = param.handle_parse_result(ctx, opts, args)
  File "c:\users\jporc\appdata\local\programs\python\python37-32\lib\site-packages\click\core.py", line 1477, in handle_parse_result
    self.callback, ctx, self, value)
  File "c:\users\jporc\appdata\local\programs\python\python37-32\lib\site-packages\click\core.py", line 96, in invoke_param_callback
    return callback(ctx, param, value)
  File "c:\users\jporc\appdata\local\programs\python\python37-32\lib\site-packages\iotedgedev\cli.py", line 497, in validate_option
[...]

PR #388, opened 2019-04-27
Fixed issue with 'cs' keyerror

jporcenaluk:
There was a key error here, and now it is using the full-blown "connectionString" as the key.

It is related to this issue: https://github.com/Azure/iotedgedev/issues/387

If 'cs' is being used by older versions of the Azure CLI, for example, there could be room here to check to see if the key exists before trying to grab it.

### DIFF
--- a/iotedgedev/azurecli.py
+++ b/iotedgedev/azurecli.py
@@ -452,7 +452,11 @@ class AzureCli:
             if result:
                 out_string = io.getvalue()
                 data = json.loads(out_string)
-                return data["connectionString"]
+                if "cs" in data:
+                    return data["cs"]
+                else:
+                    return data["connectionString"]
+
         return ''
 
     def edge_device_exists(self, value, iothub, resource_group):
@@ -495,6 +499,9 @@ class AzureCli:
             if result:
                 out_string = io.getvalue()
                 data = json.loads(out_string)
-                return data["connectionString"]
+                if "cs" in data:
+                    return data["cs"]
+                else:
+                    return data["connectionString"]
 
         return ''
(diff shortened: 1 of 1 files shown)


jongio:
@blackchoey - Can you investigate the CI build issue?  Also with this PR we'll need to figure out when the CLI introduced this change and ping iotedgedev to that min version of the CLI.

iotedgedev/azurecli.py
```diff
@@ -452,7 +452,7 @@ def get_iothub_connection_string(self, value, resource_group):
             if result:
                 out_string = io.getvalue()
                 data = json.loads(out_string)
-                return data["cs"]
+                return data["connectionString"]
```
blackchoey: Could we check both `cs` and `connectionString`. Then returns the one that exists? In this way we can support both old and new Azure CLI versions.

iotedgedev/azurecli.py
```diff
@@ -495,6 +495,6 @@ def get_device_connection_string(self, value, iothub, resource_group):
...
```
blackchoey: Same as above

blackchoey:
Hi @jporcenaluk , thanks for your contribution and sorry for the delayed response. I'm busy with another thing recently. I left some comments on the changes, could you help update the PR?

blackchoey:
@jongio My preference is to check both `cs` and `connectionString` property and return the one that exists. In this way we don't need to care about dependency versions.

blackchoey:
@jporcenaluk , I pushed some changes to support the old `cs` property. Please let me know if you have any comments.

blackchoey:
@jporcenaluk The PR has been merged. Thanks again for your contribution.

PR #391, opened 2019-05-22
Support relative path for module placeholder

blackchoey:
1. Support relative path
2. Build images when user filled the actual url in module image setting


iotedgedev/modules.py
```diff
@@ -65,7 +66,7 @@ def add(self, name, template, group_id):
 
             shutil.move(temp_template_dir, os.path.join(cwd, name))
 
-            module = Module(self.envvars, self.utility, name)
+            module = Module(self.envvars, self.utility, os.path.join(self.envvars.MODULES_PATH, name))
```
adashen: the MODULES_PATH env only used for c module? others use cwd?

iotedgedev/modules.py
```diff
@@ -138,32 +138,25 @@ def build_push(self, template_file, default_platform, no_build=False, no_push=Fa
         # sample: 'localhost:5000/filtermodule:0.0.1-amd64'
         tags_to_build = set()
 
+        deployment_manifest = DeploymentManifest(self.envvars, self.output, self.utility, template_file, True)
+
+        # get image tags for ${MODULES.modulename.xxx} placeholder
         if os.path.isdir(self.envvars.MODULES_PATH):
             for module_name in os.listdir(self.envvars.MODULES_PATH):
```
adashen: so it is still use module folder name as module name?

iotedgedev/modules.py
```diff
@@ -239,6 +232,25 @@ def build_push(self, template_file, default_platform, no_build=False, no_push=Fa
...
+                tag = "{0}:{1}{2}-{3}".format(module.repository.lower(), module.tag_version, container_tag, platform)
+                placeholder_tag_map["${{{0}.{1}}}".format(placeholder_base, platform)] = tag
+                placeholder_tag_map[tag] = tag
+
+                if platform == default_platform:
+                    placeholder_tag_map["${{{0}}}".format(placeholder_base)] = tag
+                elif platform == default_platform + ".debug":
+                    placeholder_tag_map["${{{0}.{1}}}".format(placeholder_base, "debug")] = tag
+
+                tag_build_profile_map[tag] = BuildProfile(module_name, dockerfile, module.context_path, module.build_options)
```
adashen: module_name here actually is not necessary. It seems only used to output some log later.

tests/assets/deployment.template_without_schema_template.json
```diff
@@ -69,6 +69,16 @@
...
               "createOptions": {}
             }
+          },
+          "sample_module_2": {
+            "version": "1.0",
+            "type": "docker",
+            "status": "running",
+            "restartPolicy": "always",
+            "settings": {
+              "image": "${MODULEDIR<./sample_module_2>}",
```
adashen: add case for escaped value like "image": "${MODULEDIR<.\\sample_module_2>}"

iotedgedev/modules.py
blackchoey: C module also works in cwd. The MODULES_PATH is a relative path which points to the modules folder under cwd. Other modules have their own cli to do scaffolding but for c we do the scaffolding manually. That's why you just see this code in C module.

iotedgedev/modules.py
blackchoey: We cannot use regex to grab module name from the ${MODULES.xxx} syntax, so we still need to assume the folder name is module name and user need to write the place holder in this format ${MODULES.FOLDER_NAME.PLATFORM}.
The BYPASS MODULES is filtering the key of modules in deployment template as we synced offline.

iotedgedev/modules.py
blackchoey: Yes, the logs will help users understand what's being built so we could keep it here.

iotedgedev/modules.py
adashen: Line 37 already assign cwd to MODULE_PATH to I think better to use cwd here instead of env

iotedgedev/envvars.py
```diff
@@ -118,7 +118,6 @@ def load(self, force=False):
                 self.get_registries()
 
                 self.BYPASS_MODULES = self.get_envvar("BYPASS_MODULES", default="")
-                self.ACTIVE_DOCKER_PLATFORMS = self.get_envvar("ACTIVE_DOCKER_PLATFORMS", altkeys=["ACTIVE_DOCKER_ARCH"], default="")
```
jongio: Adding note for backwards compat issues when removing this.

blackchoey: Thanks, we will add notes for this change.

PR #400, opened 2019-07-09
allow non module project 

### LINKED ISSUE #396: Azure IOT Edge build module images task fails to build after being automatically updated to 2.0.0
- iotedgedev Version: 2.0.0
- Python Version: 2.7
- Pip Version: 8.1.1 or 19.1.1
- Development machine OS Version: Hosted Ubuntu 1604 in Azure Devops
- IoT Edge device OS Version:

Steps to Reproduce:

1. create a build definition in azure devops and enable continuous integration on master or any branch you need to build
2. add an Azure Iot Edge task with action "Build module images", default platform of amd64 and deployment.template.json as template json file

I get an error

`
ERROR: No module.json file found. module.json file is required in the root of your module folder
Error: The process '/usr/local/bin/iotedgedev' failed with exit code 1
`

with versions 2.2.0 and 2.0.0 of Azure IoT Edge task and iotedgedev, respectively.

When iotedgedev and Azure IoT Edge task versions were 1.1.0 and 2.0.7 (as of 11 days ago), builds were successful.

blackchoey:
Fix #396 
1. Allow non module project.
2. Remove language property which is not used in IoT Edge development now. Some customers removed this property in module.json, so reading this property may cause errors.
3. Fix unicode error in python2 due to recent commentjson update

### DIFF
--- a/iotedgedev/modules.py
+++ b/iotedgedev/modules.py
@@ -113,6 +113,7 @@ class Modules:
             deployment_manifest_debug = DeploymentManifest(self.envvars, self.output, self.utility, self.envvars.DEPLOYMENT_CONFIG_DEBUG_TEMPLATE_FILE, True)
             deployment_manifest_debug.add_module_template(name, debug_create_options, True)
             deployment_manifest_debug.dump()
+            
         self._update_launch_json(name, template, group_id)
 
         self.output.footer("ADD COMPLETE")
--- a/iotedgedev/modules.py
+++ b/iotedgedev/modules.py
@@ -113,7 +113,7 @@ class Modules:
             deployment_manifest_debug = DeploymentManifest(self.envvars, self.output, self.utility, self.envvars.DEPLOYMENT_CONFIG_DEBUG_TEMPLATE_FILE, True)
             deployment_manifest_debug.add_module_template(name, debug_create_options, True)
             deployment_manifest_debug.dump()
-
+    
         self._update_launch_json(name, template, group_id)
 
         self.output.footer("ADD COMPLETE")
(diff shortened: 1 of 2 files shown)


iotedgedev/modules.py
```diff
@@ -145,8 +145,11 @@ def build_push(self, template_file, default_platform, no_build=False, no_push=Fa
         modules_path = os.path.join(template_file_folder, self.envvars.MODULES_PATH)
         if os.path.isdir(modules_path):
             for folder_name in os.listdir(modules_path):
-                module = Module(self.envvars, self.utility, os.path.join(modules_path, folder_name))
-                self._update_module_maps("MODULES.{0}".format(folder_name), module, placeholder_tag_map, tag_build_profile_map, default_platform)
+                try:
+                    module = Module(self.envvars, self.utility, os.path.join(modules_path, folder_name))
+                    self._update_module_maps("MODULES.{0}".format(folder_name), module, placeholder_tag_map, tag_build_profile_map, default_platform)
+                except FileNotFoundError:
```
adashen: we also need handle even module.json is found but the schema is not correct. I think it should not impact the processing of other modules.

iotedgedev/modules.py
```diff
@@ -301,7 +304,10 @@ def _get_launch_json(self, name, template, group_id):
             launch_json_content = Utility.get_file_contents(launch_json_file)
             for key, value in replacements.items():
                 launch_json_content = launch_json_content.replace(key, value)
-            launch_json = commentjson.loads(launch_json_content)
+            if PY2:
+                launch_json = commentjson.loads(launch_json_content.encode('ascii'))
```
adashen: what is the default encoding? utf-8 cannot work?

blackchoey: Since commentjson v0.7.2, we get "decoding Unicode is not supported" error on python 2.7. Other people also met this issue and I also shared our situation in the issue: https://github.com/vaidik/commentjson/issues/23
I haven't hear from the owner yet. Though limit the version of commentjson to < 0.7.2 can resolve this issue, I added this temporary solution to avoid possible dependency issue around the package.

iotedgedev/modules.py
blackchoey: Current behavior should aligns with the behavior in v1.x version. What's the expected behavior when the schema is not correct?

blackchoey: As discussed, just check the file path instead of catching exception for performance. And we will keep current behavior for now.

PR #402, opened 2019-07-31
do not use lark-parser enabled commentjson and bump version to 2.0.2

blackchoey:
The latest [lark-parser](https://pypi.org/project/lark-parser/) release breaks commentjson. Revert the version of commentjson to old version which doesn't use lark-parser.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,7 +60,7 @@ test_requirements = [
 
 setup(
     name='iotedgedev',
-    version='2.0.1',
+    version='2.0.2',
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -45,7 +45,7 @@ def run(self):
     'iotedgehubdev >= 0.8.0',
     'six',
     'applicationinsights < 0.11.8',
-    'commentjson',
+    'commentjson >= 0.7.2, < 0.8.0',
```
adashen: ==0.7.2?

PR #406, opened 2019-09-19
Validate deployment template/manifest

blackchoey:
1. Tries to detect as much errors as possible in the deployment file
2. By default only validation failure logs will be printed, the command won't fail when validation fails
3. Add option to fail the command if there're validation errors


iotedgedev/deploymentmanifest.py
```diff
@@ -147,3 +186,74 @@ def _get_module_content(self):
             return self.json["moduleContent"]
         else:
             raise KeyError("modulesContent")
+
+    def _validate_deployment_manifest_schema(self):
+        validation_success = True
+        try:
+            self.output.info("Validating schema of deployment manifest.")
+            deployment_schema = json.loads(urlopen(Constants.deployment_manifest_schema_url).read())
+            validator_class = jsonschema.validators.validator_for(deployment_schema)
```
adashen: cannot use jsonschema.validate() method?

iotedgedev/deploymentmanifest.py
```diff
@@ -147,3 +186,74 @@ def _get_module_content(self):
...
+        self.output.info("Start validating createOptions for all modules.")
+        modules = self.get_all_modules()
+        validation_success = True
+        for module_name, module_info in modules.items():
+            try:
+                self.output.info("Validating createOptions for module %s" % module_name)
+                create_options = []
+                if "settings" in module_info and "createOptions" in module_info["settings"]:
+                    create_options_value = module_info["settings"]["createOptions"]
+                    if len(create_options_value) > TWIN_VALUE_MAX_SIZE:
```
adashen: doesn't check whether is json object before checking len?

iotedgedev/deploymentmanifest.py
```diff
@@ -147,3 +186,74 @@ def _get_module_content(self):
             return self.json["moduleContent"]
         else:
             raise KeyError("modulesContent")
+
+    def _validate_deployment_manifest_schema(self):
```
adashen: the method has lots of duplication with validate_deployment_template. Can we refactor?

iotedgedev/deploymentmanifest.py
blackchoey: If there're multiple schema errors, jsonschema.validate() only returns 1 error. So user could not know remaining errors and fix them.
The validator way could give us all schema errors, so I implemented in this way.

iotedgedev/deploymentmanifest.py
blackchoey: createOptions may be split so it can be an invalid json string
The function will merge all createOptions and check whether it's a valid json string later: [Line 240](https://github.com/Azure/iotedgedev/blob/chyuan/deployment-manifest-validation/iotedgedev/deploymentmanifest.py#L240)

blackchoey: And since the schema defines createOptions as string, the validation relies on the schema to do the check

iotedgedev/deploymentmanifest.py
```diff
@@ -147,3 +186,74 @@ def _get_module_content(self):
...
+                    create_options.append(create_options_value)
+                    # Merge additional create options
+                    for i in range(1, TWIN_VALUE_MAX_CHUNKS):
+                        property_name = "createOptions0%d" % i
+                        if property_name in module_info["settings"]:
+                            create_options_value = module_info["settings"][property_name]
+                            if len(create_options_value) > TWIN_VALUE_MAX_SIZE:
+                                validation_success = False
+                                self.output.info("Length of %s in module %s exceeds %d" % (property_name, module_name, TWIN_VALUE_MAX_SIZE))
+                            create_options.append(create_options_value)
```
adashen: if exceed still continue top append?

iotedgedev/deploymentmanifest.py
```diff
@@ -147,3 +186,74 @@ def _get_module_content(self):
...
+                    if not create_options_string.startswith('{'):
+                        validation_success = False
+                        self.output.info("createOptions of module %s should be an object" % module_name)
+                    else:
+                        try:
+                            json.loads(create_options_string)
+                            self.output.info("createOptions of module %s validation passed" % module_name)
+                        except ValueError as err:
+                            validation_success = False
+                            self.output.info("createOptions of module %s is not a valid JSON string. Error: %s" % (module_name, err))
```
adashen: can we output warning instead of info?

iotedgedev/deploymentmanifest.py
blackchoey: Yes, so we can check whether the string is still a valid json. Though the user need to split the createOptions later.
This is for following scenario: If the createOptions exceeds max length and also is not a valid json, we could report 2 errors to user instead of telling user max length exceeded error first and then tell user not valid json error in the second run.

iotedgedev/deploymentmanifest.py
blackchoey: Actually there're only 2 output channels: stdout and stderr. Currently I outputs the validation failures to stdout channel because Azure Pipelines force print messages in stderr in build summary and there's option to fail the task if any messages arrive stderr channel.

iotedgedev/deploymentmanifest.py
blackchoey: Refactored

iotedgedev/deploymentmanifest.py
adashen: "Warning" does not mean the output channel. Instead, we should have a function in output class to specify a kind of warning message. Like [Warning] prefix

iotedgedev/deploymentmanifest.py
```diff
@@ -256,5 +230,49 @@ def _validate_createOptions(self):
         if (validation_success):
             self.output.info("Validation for all createOptions passed.")
         else:
-            self.output.info("Errors found during createOptions validation. Please check the logs for details.")
+            self.output.warning("Errors found during createOptions validation. Please check the logs for details.")
+        return validation_success
+
+    def _validate_createOptions_lengh(self, module_name, module_info):
```
jongio: Consider making the _validate_create method names consistent.  Recommend renaming this to `_validate_create_options_length`

iotedgedev/deploymentmanifest.py
```diff
@@ -216,37 +216,11 @@ def _validate_createOptions(self):
...
-                        self.output.info("createOptions of module %s should be an object" % module_name)
-                    else:
-                        try:
-                            json.loads(create_options_string)
-                            if current_module_validation_success:
-                                self.output.info("createOptions of module %s validation passed" % module_name)
-                        except ValueError as err:
-                            current_module_validation_success = False
-                            self.output.info("createOptions of module %s is not a valid JSON string. Error: %s" % (module_name, err))
+                    current_module_validation_success &= self._validate_createOptions_lengh(module_name, module_info)
```
jongio: Consider creating a method called `_validate_create_options` that does both length and format validation.  Basically a helper around the two methods.

tests/test_iotedgedev.py
```diff
@@ -481,27 +481,32 @@ def test_validate_deployment_template_and_manifest_failed():
         else:
             result = runner_invoke(['genconfig', '-f', 'deployment.template_invalidresult.json'])
 
+        template_path = os.path.join(tests_assets_dir, "deployment.template_invalidresult.json")
```
jongio: Seems like you should put this file name in a const or make it configurable.

blackchoey: Declared the file name as local variable, in case the test case need to use another test file in the future.

iotedgedev/deploymentmanifest.py
blackchoey: updated

iotedgedev/deploymentmanifest.py
blackchoey: Updated

PR #411, opened 2019-10-18
Pull image from mcr instead of docker hub

blackchoey:


README.md
```diff
@@ -39,19 +39,19 @@ The only thing you need to install is Docker. All of the other dev dependencies
     **Windows**
     ```
     mkdir c:\temp\iotedge
-    docker run -ti -v /var/run/docker.sock:/var/run/docker.sock -v c:/temp/iotedge:/home/iotedge microsoft/iotedgedev
+    docker run -ti -v /var/run/docker.sock:/var/run/docker.sock -v c:/temp/iotedge:/home/iotedge mcr.microsoft.com/iotedge/iotedgedev
```
jongio: What’s the reason for making this change? It’s cleaner and easier to remember when using dockerhub

blackchoey: We found new tags are not available on Docker Hub (we do updated related metadata files). After talked with MCR team, new tags pushed to MCR will not be synced to Docker Hub. Only existing tags on Docker Hub will be updated in case users still pulls from Docker Hub. In the meanwhile, MCR is the recommended place to pull images from. So update the documents to use MCR.

blackchoey: @jongio Per offline discussion, we will pull from MCR for now.

blackchoey:
CI failed due to known windows and macos agent issue. Does not block the merge.

Issue #424, opened 2020-03-11
Some deployment manifest validation error don't fail genconfig command


blackchoey:
Example
`"test####":"test"` in settings
`"test 123":"test 1234"` in properties.desired

PR #425, opened 2020-03-13
Fix genconfig does not fail when schema validation failed

blackchoey:
Fix genconfig does not fail when schema validation failed #424 
AB#1696447

### DIFF
--- a/iotedgedev/deploymentmanifest.py
+++ b/iotedgedev/deploymentmanifest.py
@@ -182,17 +182,14 @@ class DeploymentManifest:
             validator_class = jsonschema.validators.validator_for(schema_object)
             validator = validator_class(schema_object)
             validation_errors = validator.iter_errors(self.json)
-            error_detected = False
             for error in validation_errors:
-                error_detected = True
-                self.output.warning("%s schema error: %s. Property path:%s" % (schema_type, error.message, "->".join(error.path)))
-            if error_detected:
-                self.output.warning("%s schema validation failed. Please see previous logs for more details" % schema_type)
-            else:
-                self.output.info("%s schema validation passed." % schema_type)
-
-            if error_detected:
                 validation_success = False
+                self.output.warning("%s schema error: %s. Property path:%s" % (schema_type, error.message, "->".join(error.path)))
+            if validation_success:
+                self.output.info("%s schema validation passed." % schema_type)
+            else:
+                self.output.warning("%s schema validation failed. Please see previous logs for more details" % schema_type)
+
         except jsonschema.exceptions.SchemaError as schemaErr:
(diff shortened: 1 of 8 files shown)


iotedgedev/deploymentmanifest.py
```diff
@@ -190,6 +190,9 @@ def _validate_json_schema(self, schema_object, json_object, schema_type):
                 self.output.warning("%s schema validation failed. Please see previous logs for more details" % schema_type)
             else:
                 self.output.info("%s schema validation passed." % schema_type)
+
+            if error_detected:
```
adashen: line 189 has if error_detected?

adashen: besides it already has error_detected, do we still need validation_success?

iotedgedev/deploymentmanifest.py
```diff
@@ -146,7 +146,7 @@ def validate_deployment_template(self):
         validation_success = True
         try:
             template_schema = json.loads(urlopen(Constants.deployment_template_schema_url).read().decode())
-            self._validate_json_schema(template_schema, self.json, "Deployment template")
+            validation_success = self._validate_json_schema(template_schema, self.json, "Deployment template")
         except Exception as ex:  # Ignore other non shcema validation errors
             self.output.info("Unexpected error during deployment template schema validation, skip schema validation. Error:%s" % ex)
```
adashen: exception happens and the validation is still success?

iotedgedev/deploymentmanifest.py
blackchoey: Removed the error_detected and reuse validation_success. We need to declare validation_success outside of try to handle situations with exception.

iotedgedev/deploymentmanifest.py
blackchoey: Only schema failures should be treated as validation failure. Other exceptions such as network error should not fail the validation. iotedgedev will only print related error and skip validation.

iotedgedev/deploymentmanifest.py
adashen: removed error_detected? we have two blocks of error_detected (189 and 194)?

blackchoey:
Some Azure CLI modules are broken, I will investigate and see whether I can provide a quick fix

iotedgedev/deploymentmanifest.py
blackchoey: Here's the commit that removes error_detected, I removed both of them and use validation_success instead: https://github.com/Azure/iotedgedev/pull/425/commits/60cac6655901b22fad8de6336616c1e28e042373

Issue #426, opened 2020-03-31
iotedgedev in Azure DevOps pipeline fails - ImportError: No module named configparser


adriandiver:
Hi,

I setup a simple IoT Edge DevOps Project through the Azure Portal (Python). The pipelines worked well until yesterday and now the pipeline fails during the step "Azure IoT Edge - Build module images". Based on the output it seems there is an issue with configparser:

##[debug]which 'iotedgedev'
##[debug]found: '/usr/local/bin/iotedgedev'
##[debug]which 'iotedgedev'
##[debug]found: '/usr/local/bin/iotedgedev'
##[debug]/usr/local/bin/iotedgedev arg: --version
##[debug]/usr/local/bin/iotedgedev arg: --version
##[debug]exec tool: /usr/local/bin/iotedgedev
##[debug]exec tool: /usr/local/bin/iotedgedev
##[debug]arguments:
##[debug]arguments:
##[debug]   --version
##[debug]   --version
##[debug]
##[error]Traceback (most recent call last):
  File "/usr/local/bin/iotedgedev", line 7, in <module>
    from iotedgedev.cli import main
  File "/usr/local/lib/python2.7/dist-packages/iotedgedev/cli.py", line 12, in <module>
    from .azurecli import AzureCli
  File "/usr/local/lib/python2.7/dist-packages/iotedgedev/azurecli.py", line 9, in <module>
    from azure.cli.core import get_default_cli
  File "/usr/local/lib/python2.7/dist-packages/azure/cli/core/__init__.py", line 24, in <module>
    from .local_context import AzCLILocalContext, STORE
  File "/usr/local/lib/python2.7/dist-packages/azure/cli/core/local_context.py", line 9, in <module>
    import configparser
ImportError: No module named configparser

I even tried to install an older version of iotedgedev but it still fails. 

Steps to Reproduce:

1. Deploy Azure DevOps project through Azure Portal (IoT Edge, Python)
2. Run the Pipeline

mrtncls:
Hello,

I want to report the same issue.

Starting from yesterday, we see that iotedgedev fails to install with message "ImportError: No module named configparser".

<details>
<summary>Task output since yesterday</summary>

```
2020-04-01T06:51:21.0303370Z ##[section]Starting: Azure IoT Edge - Generate deployment manifest
2020-04-01T06:51:21.0311591Z ==============================================================================
2020-04-01T06:51:21.0311949Z Task         : Azure IoT Edge
[...]

PR #427, opened 2020-04-01
Fix error caused by latest azure cli core version

blackchoey:
The latest azure cli core version only works on python3, revert to previous version to fix errors on python2


blackchoey:
Fix #426

PR #442, opened 2020-12-19
remove extraneous cryptography package, remove static jdk env var

smcd253:
Allows hosted ubuntu 1604 to complete pip install and run tests
9/100 tests still fail
* connection issues (maybe due to extraction of cryptography?)
* edge simulator started in simulation mode...


blackchoey (approved):

smcd253:
Update: pipeline runs manually with proper iot hub env vars and container registry env var pointed to localhost:5000

PR #447, opened 2021-01-08
Remove support for Python 2

smcd253:
- Remove support for Python2... only Python 3.6, 3.7, and 3.8 are currently supported
- Add testing to validate version catch for versions less than Python 3.0


iotedgedev/envvars.py
```diff
@@ -142,8 +121,6 @@ def load(self, force=False):
                 "Variable that caused exception: {0}".format(str(ex))
                 raise ValueError(msg)
 
-        self.clean()
```
marianan, line 145: >self.clean() [](start = 8, length = 12)

why removing the clean here?

iotedgedev/iothub.py
```diff
@@ -1,4 +1,4 @@
-from .compat import PY35
+from .version import PY35
```
marianan, line 1: >PY35 [](start = 21, length = 4)

PY35 is defined in compat.py, which is removed. How is this import resolving?

iotedgedev/utility.py
```diff
@@ -93,7 +89,7 @@ def get_file_contents(file, expandvars=False):
                 return content
 
     def decode(self, val):
-        return val.decode("utf-8").strip()
+        return val.decode().strip()
```
marianan, line 92: do we no longer need to pass the encoding as a parameter?

setup.py
```diff
@@ -62,14 +60,12 @@
     license='MIT license',
     zip_safe=False,
     keywords='azure iot edge dev tool',
-    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, <3.8',
+    python_requires='!=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, <3.8',
```
marianan, line 63: >!=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.* [](start = 21, length = 52)

Will this be equivalent - '>=3.6'?

iotedgedev/envvars.py
smcd253: Correct me if I'm wrong, but the implementation above was for python 2 only, so it serves no purpose for python 3.

iotedgedev/utility.py
smcd253: no. python 3 defaults to utf-8

setup.py
smcd253: yes indeed! you are right. 
It should be `python_requires = '>=3.6, <3.8'`

marianan (approved):
:shipit:

jongio:
Would be good if @digimaun could review too

setup.py
```diff
@@ -62,14 +60,12 @@
     license='MIT license',
     zip_safe=False,
     keywords='azure iot edge dev tool',
-    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, <3.8',
+    python_requires='>=3.6, <3.8',
```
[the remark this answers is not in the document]
smcd253: haven't tested with >3.8. Happy to add in a pipeline testing the latest version in a separate PR, but wanted to minimize the scope of this PR as it is already quite large.

tests/test_buildoptionsparser.py
```diff
@@ -1,9 +1,13 @@
 import pytest
-from iotedgedev.buildoptionsparser import BuildOptionsParser
 
-pytestmark = pytest.mark.unit
+from .version import test_py2, minversion
+
+if not test_py2:
```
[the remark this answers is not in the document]
smcd253: Many libraries are py3 specific. This prevents tests from failing and allows me to raise an expected assertion error on the version catch in iotedgedev/version_catch.py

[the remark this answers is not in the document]
digimaun: I don't quite understand this comment, can you please expand on what is failing? 

By updating the `python_requires` constraint pip should only install the package in the supported range and error out otherwise. Then you can apply version or env constraints to package dependencies or requirements file entries if needed.

It would be good to use as much of the available existing tooling as possible.

tests/__init__.py
```diff
@@ -1,5 +1,3 @@
-# -*- coding: utf-8 -*-
-
```
digimaun, line 2: Maybe replace `__auther__` , `__email__` to team identifiers/email if you want to keep these attributes.

README.md
```diff
@@ -166,6 +166,9 @@ The only thing you need to install is Docker. All of the other dev dependencies
 
     </details>
 
+## Python Versions
+This project only supports Python 3.6 and 3.7.
```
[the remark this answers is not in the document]
smcd253: 3.7 and 3.8. Will make changes now and will make it 3.7+ when we've tested with the latest version.

tests/test_buildoptionsparser.py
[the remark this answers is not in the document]
smcd253: The idea was that even if pip install fails, tox and pytest will be installed and can still run. This solution provides a catch for testing, avoiding all tests except for the e2e test I built to get to the version catch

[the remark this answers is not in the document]
digimaun: OK I see. Testing infrastructure wise (like via tox) you should be able to configure tox to properly install the package in the defined compatible python runtimes and abort the run if a failure occurs in package installation. Then philosophically, this package defines its python run requirements via attributes like `classifiers` and `python_requires`, which is a standard approach for the Python ecosystem and the Python ecosystem tools + users should abide. In general, if the package claims its 3.6+ and a user tries to run aspects in a non-compatible environment then failure is expected.

That being said, to still achieve the goal here, but in less code consider refactoring to use one of the following pytest configuration approaches:
* [collect_ignore_glob or collect_ignore](https://docs.pytest.org/en/latest/reference.html#globalvar-collect_ignore_glob)
* A wider scoped fixture...something like

```python
@pytest.fixture(scope="session", autouse=True)
def ensure_python_version():
    if sys.version_info < (3, 6):
        pytest.skip("Incompatible environment. The package requires Python 3.6+.")
```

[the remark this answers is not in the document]
smcd253: This is exactly the kind of thing I was looking for! I wanted a global way to skip tests but could not find one. 

I will investigate configuring tox properly first and will use this as backup.

Thanks @digimaun!

smcd253:
Closing and issuing new PR for Python 2 removal without testing.

PR #449, opened 2021-01-26
Remove Python 2 Support

smcd253:
Remove support for Python 2... Only works with Python 3.6 and 3.7


setup.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.2.0
+current_version = 3.0.0
```
marianan, line 2: >3.0.0 [](start = 18, length = 5)

Please release -rc version first, according to OneNote instructions. We will use the RC version for testing before releasing the major version. To release a new version the version reported from code needs to match with the git tag.

setup.py
```diff
@@ -45,7 +43,7 @@
 
 setup(
     name='iotedgedev',
-    version='2.2.0',
+    version='3.0.0',
```
marianan, line 46: >version='3.0.0 [](start = 4, length = 14)

same comment about releasing an rc version first. The OneNote has isntructions for all the places in the code that need to be updated for the RC version release to go through.

setup.py
```diff
@@ -62,14 +60,12 @@
     license='MIT license',
     zip_safe=False,
     keywords='azure iot edge dev tool',
-    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, <3.8',
+    python_requires='!=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, <3.8',
```
marianan, line 63: >!=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.* [](start = 21, length = 52)

can we replace it with '>=3.6'?

marianan (changes requested):
:clock1:

setup.py
smcd253: ah right, yes. meant to bring this in from the other PR. thanks for catching this.

setup.cfg
smcd253: I thought the -rc was only done on the git tag and not in the codebase.

CHANGELOG.md
```diff
@@ -1,10 +1,9 @@
 # Changelog
 All notable changes to this project since 0.82.0 will be documented in this file.
 
-## [2.2.0] - 2021-01-06
-### Changed
-- Remove explicit cryptography package install to resolve dependency conflicts so tests can run
-- Add local registry support for Windows
+## [3.0.0] - 2021-01-04
+- Remove support for Python2... only Python 3.6 and 3.8 are currently supported
```
jongio, line 5:
```suggestion
- Remove support for Python2. Only Python 3.6 and 3.8 are currently supported
```

iotedgedev/__init__.py
```diff
@@ -1,8 +1,6 @@
-# -*- coding: utf-8 -*-
-
 """Top-level package for Azure IoT Edge Dev Tool."""
 
-__author__ = 'Microsoft Corporation'
-__email__ = 'vsciet@microsoft.com'
-__version__ = '2.2.0'
+__author__ = 'Microsoft Open Source'
+__email__ = 'opencode@microsoft.com'
```
jongio, line 4: Is it a standard to use this alias?  Who monitors it?  The old alias was the team, not general alias.

jongio (approved):
My approval is conditional on @digimaun 's approval.

iotedgedev/__init__.py
smcd253: This is how the sdk teams do it. I don't believe we have a team email. The emails get filtered down to the team owning the repo.

avagraw (approved):

marianan (approved):
:shipit:

CHANGELOG.md
```diff
@@ -1,10 +1,9 @@
 # Changelog
 All notable changes to this project since 0.82.0 will be documented in this file.
 
-## [2.2.0] - 2021-01-06
-### Changed
-- Remove explicit cryptography package install to resolve dependency conflicts so tests can run
-- Add local registry support for Windows
+## [3.0.0] - 2021-01-04
+- Remove support for Python2. Only Python 3.6 and 3.8 are currently supported
```
digimaun, line 5: Only python 3.6 and 3.8? Why not Python 3.7?

setup.py
```diff
@@ -62,14 +60,12 @@
     license='MIT license',
     zip_safe=False,
     keywords='azure iot edge dev tool',
-    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, <3.8',
+    python_requires='>=3.6, <3.8',
```
digimaun, line 63: Why lock out 3.8?

setup.py
```diff
@@ -62,14 +60,12 @@
...
     classifiers=[
         'Development Status :: 2 - Pre-Alpha',
         'Intended Audience :: Developers',
         'License :: OSI Approved :: MIT License',
         'Natural Language :: English',
-        'Programming Language :: Python :: 2',
-        'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7'
```
digimaun, line 71: Add the Python 3.8 metadata classifier if 3.8 is supposed to be supported.

CHANGELOG.md
smcd253: Woops, typo. Supposed to be 3.7 and 3.8

setup.py
smcd253: have not tested with 3.8 or 3.9. Will do this in a separate PR, as it is out of scope of this one.

tests/__init__.py
```diff
@@ -1,7 +1,5 @@
-# -*- coding: utf-8 -*-
-
 """Test package for Azure IoT Edge Dev Tool ."""
 
-__author__ = 'David Ruttka'
-__email__ = 'david.ruttka@microsoft.com'
-__version__ = '0.0.1'
+__author__ = 'Microsoft Open Source'
```
digimaun, line 3: I don't think these attributes are that useful in tests unless you foresee lots of community involvement here.

setup.py
marianan: we have not tested 3.8 yet so cannot confirm support just yet.

tests/__init__.py
smcd253: This is consistent with our other client repos (like the python sdk).

setup.py
smcd253: Sorry, that was a typo. Will add this in in a separate PR when it has been tested.

iotedgedev/utility.py
```diff
@@ -93,7 +89,7 @@ def get_file_contents(file, expandvars=False):
                 return content
 
     def decode(self, val):
-        return val.decode("utf-8").strip()
```
digimaun, line 96: nit, though utf-8 may be default its nice to explicitly keep the encoding args because the next set of eyes  quickly see understand with less knowledge.

iotedgedev/version.py
```diff
@@ -0,0 +1,6 @@
+import sys
+
+PY35 = sys.version_info >= (3, 5)
```
digimaun, line 3: If this check is supposed to be consistent across `iotedgedev`, you could put it in `iotedgedev/__init__.py` so that it is evaluated upon package load.

smcd253: I tested this and it has no effect on the time it takes to throw an error. Furthermore, in neither scenario is the assert actually thrown, which is interesting because I got the desired output with the tests...

smcd253: I'm going to keep it as is, as it has no affect on performance or error-throw time.

digimaun: This comment wasn't aimed for performance - it was aimed to reduce code redundancy.

digimaun: And if the expected behavior  is not working, seems like it would be good to investigate and fix/refactor.

digimaun (approved):

iotedgedev/version.py
smcd253: Fair enough and good point. I'll find an appropriate solution.

smcd253: Confirmed that the version assert only provided value for testing. It has been removed.

PR #466, opened 2021-03-18
update iotedgedev to support Edge runtime 1.1

cindydeng1998:
- added -rt flag during iotedgedev init -rt [version_number] to specify a version of runtime (default 1.1)
- Validated against Raspi to update edgeHub and edgeAgent images to 1.1
- changed all template deployment.json files to take in dynamic ${RUNTIME_TAG}


docker/runtime/arm/rund.arm32v7.sh
```diff
@@ -3,7 +3,7 @@
 echo '=> detecting IP'
 export IP=$(ip addr show eth0 | grep "inet\b" | awk '{print $2}' | cut -d/ -f1)
 export IOT_DEVICE_HOSTNAME="$IP"
-export AGENT_IMAGE="mcr.microsoft.com/azureiotedge-agent:1.0@sha256:3fdb80a6dfe1fbcbf11cfcea56dda79d1d779490c66d394a2b3f388d74ab0c26"
+export AGENT_IMAGE="mcr.microsoft.com/azureiotedge-agent:1.1@sha256:84d81e0527799e903fca9602c56efcbc43ce6f90eaeb7316fddc7d663d40d1b1"
```
marianan, line 6: >1.1 [](start = 57, length = 3)

Can we parametrize the values here?

iotedgedev/cli.py
```diff
@@ -67,12 +67,18 @@ def docker():
                        "Use \".\" as NAME to create in the current folder.")
 @click.argument("name",
                 required=True)
+@click.option("--runtime_tag",
+              "-rt",
```
marianan, line 71: >-rt [](start = 15, length = 3)

Could you sync with Haitham on the namign of the parameter, he is enabling this for the simualtor and it may be more intuitive for users if they see similar parameter name.

iotedgedev/envvars.py
```diff
@@ -98,7 +98,7 @@ def load(self, force=False):
 
                 self.BYPASS_MODULES = self.get_envvar("BYPASS_MODULES", default="")
                 self.CONTAINER_TAG = self.get_envvar("CONTAINER_TAG", default="")
-                self.RUNTIME_TAG = self.get_envvar("RUNTIME_TAG", default="1.0")
+                self.RUNTIME_TAG = self.get_envvar("RUNTIME_TAG", default="")
```
marianan, line 101: >"" [](start = 74, length = 2)

Why empty default value?

tests/test_deploymentmanifest.py
```diff
@@ -71,8 +71,10 @@ def test_get_all_modules(deployment_manifest):
 def test_add_module_template(deployment_manifest):
     deployment_manifest = deployment_manifest(test_file_1)
     deployment_manifest.add_module_template("csharpmodule2")
-    with open(test_file_3, "r") as expected:
-        assert deployment_manifest.json == json.load(expected)
+    expected = json.loads(Utility.get_file_contents(test_file_3, expandvars=True))
+    assert deployment_manifest.json == expected
+    #with open(test_file_3, "r") as expected:
+     #   assert deployment_manifest.json == json.load(expected)
```
marianan, line 77: is the commented code work in progress?

tests/assets/deployment.manifest_invalid.json
```diff
@@ -25,15 +25,15 @@
             "edgeAgent": {
               "type": "docker",
               "settings": {
-                "image": "mcr.microsoft.com/azureiotedge-agent:1.0"
+                "image": "mcr.microsoft.com/azureiotedge-agent:${RUNTIME_TAG}"
```
marianan, line 28: >${RUNTIME_TAG} [](start = 63, length = 14)

Is the manifest invalid because the env variable isn't sepcified?

marianan (changes requested):
:clock1:

tests/assets/deployment.manifest_invalid.json
cindydeng1998: I believe this file is being used to test against a user generated manifest that is invalid somewhere, not because of the .env variable.

tests/test_deploymentmanifest.py
cindydeng1998: The commented code is old code, not work in progress. Forgot to remove :)

iotedgedev/envvars.py
cindydeng1998: It has to do with the sequence of calls that creates the template files, which you want to leave as $(RUNTIME_TAG) when they're being generated to be swapped later down the line. The default is being specified somewhere else.

docker/runtime/arm/rund.arm32v7.sh
cindydeng1998: Still investigating the local docker iotedgedev arm situation, we'll need to update the SHAs along with the runtime version which we might also want to be generated dynamically

marianan, line 6: >1.1 [](start = 57, length = 3)

I think we should revert the runtime images version to 1.0 in the docker files.

iotedgedev/cli.py
```diff
@@ -67,12 +67,18 @@ def docker():
...
+@click.option("--edge-runtime-version",
+              "-er",
+              default="1.0",
+              show_default=True,
+              required=False,
+              help="Specify the IoT Edge Runtime Version. Currently available 1.0 and 1.1")
 @add_module_options(envvars, init=True)
 @with_telemetry
-def new(name, module, template, group_id):
+def new(name, module, template, edge-runtime-version, group_id):
```
marianan, line 78: >edge-runtime-version [](start = 32, length = 20)

I think we should allow only 1.0 or 1.1 for the runtime versions and throw an error if the value is not within that list.

cindydeng1998: resolved in new commit https://github.com/Azure/iotedgedev/pull/466/commits/69998154b99df5ac7615e1f4d5b943ce6fb3c9cf

thanks for catching it!

docker/runtime/arm/rund.arm32v7.sh
cindydeng1998: reverted back in commit https://github.com/Azure/iotedgedev/pull/466/commits/4f2deafc933ae18895d3c17e1dacc8b47c89b5a3

iotedgedev/cli.py
```diff
@@ -82,15 +88,25 @@ def new(name, module, template, group_id):
...
+              show_default=True,
+              required=False,
+              help="Specify the IoT Edge Runtime Version. Currently available 1.0 and 1.1")
 @add_module_options(envvars, init=True)
 @with_telemetry
-def init(module, template, group_id):
+def init(module, template, group_id, edge_runtime_version):
+    if edge_runtime_version is not None:
+        if (edge_runtime_version != "1.0" or edge_runtime_version != "1.1"):
+            output.info('-edge-runtime-version `{0}` is not valid. Currently support versions 1.0 and 1.1 only'.format(edge_runtime_version))
```
mhshami01, line 102: should this be 'Currently supported versions are ...'?

marianan (approved):
:shipit:

PR #468, opened 2021-04-09
Update setuptools so az cli can be used within the Windows devops pipeline

marianan:
This is a work around an az cli issue. All iotedgedev pipelines pass with this fix.


cindydeng1998 (approved):
🚢

PR #478, opened 2021-06-08
Push iotedgedev container to MCR

marianan:
MCR push process is already described at "MCR push process" OneNote. This change will make sure to not push to DockerHub but instead to MCR.

### DIFF
--- a/docker/tool/push-docker.sh
+++ b/docker/tool/push-docker.sh
@@ -27,6 +27,7 @@ function show_help
 IMAGE_NAME="$1"
 PLATFORM="$2"
 VERSION="$3"
+ACR_LOGIN_SERVER="iotedgetoolscontainerregistry.azurecr.io"
 
 if [ "$IMAGE_NAME" = "--help" ]; then    
     echo "Usage:"
@@ -40,19 +41,19 @@ if [ -z "$VERSION" ]; then
     echo "Detected version $VERSION"
 fi
 
-docker push $IMAGE_NAME:$VERSION-amd64 
-docker push $IMAGE_NAME:latest-amd64 
+docker push $ACR_LOGIN_SERVER/$IMAGE_NAME:$VERSION-amd64 
+docker push $ACR_LOGIN_SERVER/$IMAGE_NAME:latest-amd64 
 
 #TODO add windows container support.  For now we only push linux.
-#docker push $IMAGE_NAME:$VERSION-windows-amd64
-#docker push $IMAGE_NAME:latest-windows-amd64
+#docker push $ACR_LOGIN_SERVER/$IMAGE_NAME:$VERSION-windows-amd64
+#docker push $ACR_LOGIN_SERVER/$IMAGE_NAME:latest-windows-amd64
 
 echo -e "\n===== Creating Multi-Arch Docker image"
 #docker manifest create $IMAGE_NAME:latest $IMAGE_NAME:latest-amd64 $IMAGE_NAME:latest-windows-amd64 
 docker manifest create --insecure $IMAGE_NAME:latest $IMAGE_NAME:latest-amd64 
 
 echo -e "\n===== Pushing Docker Multi-Arch image"
-docker manifest push --purge $IMAGE_NAME:latest
+docker manifest push --purge $ACR_LOGIN_SERVER/$IMAGE_NAME:latest
 
 if [ in_docker_folder = 0 ]; then
     cd original_folder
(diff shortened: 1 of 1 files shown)


docker/tool/push-docker.sh
```diff
@@ -1,59 +1,60 @@
...
-echo -e "\n===== Creating Multi-Arch Docker image"
-#docker manifest create $IMAGE_NAME:latest $IMAGE_NAME:latest-amd64 $IMAGE_NAME:latest-windows-amd64 
-docker manifest create --insecure $IMAGE_NAME:latest $IMAGE_NAME:latest-amd64 
-
-echo -e "\n===== Pushing Docker Multi-Arch image"
-docker manifest push --purge $IMAGE_NAME:latest
-
-if [ in_docker_folder = 0 ]; then
-    cd original_folder
+#!/bin/bash
```
mhshami01, line 1: did we change line endings or something? was it intentional?

avagraw (approved):

docker/tool/push-docker.sh
marianan: there is no diff in this line, will connect offline to udnerstand your question.

cindydeng1998:
It doesn't look like this file has been edited in 3 years, and we have manual steps in OneNote to push images. Is the process now to use this script instead of the OneNote documentation?

docker/tool/push-docker.sh
mhshami01: on master:
image
on your branch:
image

mhshami01: as you can see, on branch user/marine/mcrpush, the file is now using windows style line endings, which not good in a bash script :-)

marianan:
i think we should use the script and make sure that it works for what we need it for. We already have some many places with instructions in documentation.

---
In reply to: [858800036](https://github.com/Azure/iotedgedev/pull/478#issuecomment-858800036) [](http://example.com/codeflow?ancestors=858800036,858800036)

docker/tool/push-docker.sh
marianan: good catch, i will update

docker/tool/push-docker.sh
```diff
@@ -1,59 +1,60 @@
...
+    in_docker_folder=1
+else
+    in_docker_folder=0
+    cd docker
+fi
+
+function show_help
+{
+    echo "Usage:"
+    echo "push-docker.sh <dockerhub> <platform> [<version>]"
```
cindydeng1998, line 19: the first entry here is dockerhub, but the first argument is IMAGE_NAME="$1" on line 27. which one should we keep?

docker/tool/push-docker.sh
```diff
@@ -1,59 +1,60 @@
...
+#!/bin/bash
+
+# stop on error
+set -e
+
+# make sure we're in docker folder
+original_folder=$PWD
+
+if [ -z $"echo $PWD | grep /docker$" ]; then 
+    in_docker_folder=1
```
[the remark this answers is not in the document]
marianan: I will remove this logic here and at the end of the file

docker/tool/push-docker.sh
```diff
@@ -1,59 +1,60 @@
...
+    echo "push-docker.sh imagename [linux|windows]"
+    exit 1
+fi
+
+if [ -z "$VERSION" ]; then
+    echo -e "\n===== Detecting version"
+    VERSION=$(cat ../iotedgedev/__init__.py | grep '__version__' | grep -oP "'\K[^']+")
+    echo "Detected version $VERSION"
+fi
+
```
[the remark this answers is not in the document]
marianan: will push these 2 containers as well:
-t $IMAGE_NAME:$VERSION \
        -t $IMAGE_NAME:latest \

docker/tool/push-docker.sh
```diff
@@ -1,60 +1,35 @@
...
+IMAGE_NAME="iotedgedev"
+VERSION="$1"
+
+if [ "$VERSION" = "--help" ]; then    
+    show_help
+fi
+
+if [ -z "$VERSION" ]; then
+    echo -e "\n===== Detecting version"
+    VERSION=$(cat ../../iotedgedev/__init__.py | grep '__version__' | grep -oP "'\K[^']+")
```
mhshami01, line 27: Could be simplified as follows...
VERSION=$(grep '__version__' ../../iotedgedev/__init__.py  | grep -oP "'\K[^']+")

mhshami01 (approved):

cindydeng1998 (approved):
🌞

mhshami01 (approved):

PR #479, opened 2021-06-15
Unblock build_docker.sh and Update Dockerfile.base

cindydeng1998:
- update node version
- remove old node dependency


marianan (approved):
:shipit:

Issue #430, opened 2020-06-04
support for layered deployment


spinosae:
Our team has adopted [layered deployment](https://docs.microsoft.com/en-us/azure/iot-edge/module-deployment-monitoring#layered-deployment) to add edge modules on top of base deployment managed by another team. It is therefore in our interest to build a layered deployment manifest only. Here is a sample from MS docs:
```
{
  "content": {
    "modulesContent": {
      "$edgeAgent": {
        "properties.desired.modules.SimulatedTemperatureSensor": {
          "settings": {
            "image": "mcr.microsoft.com/azureiotedge-simulated-temperature-sensor:1.0",
              "createOptions": ""
          },
          "type": "docker",
          "status": "running",
          "restartPolicy": "always",
          "version": "1.0"
        }
      },
      "$edgeHub": {
        "properties.desired.routes.upstream": "FROM /messages/* INTO $upstream"
      },
      "SimulatedTemperatureSensor": {
        "properties.desired": {
          "SendData": true,
          "SendInterval": 5
        }
      }
    }
  }
}
```
This format is not recognized by iotedgedev and would fail to build. Do you have plan to support it in your road map?

blackchoey:
Hi @spinosae You need a full IoT Edge deployment template in order to build your solutions, which tells the tool which module needs to be built.
The layered deployment is very likely to miss these information, so couldn't be used to build your solutions. However, you can always commit this layered deployment to your repo along with the traditional deployment manifest. This won't impact the build process.

spinosae:
> Hi @spinosae You need a full IoT Edge deployment template in order to build your solutions, which tells the tool which module needs to be built.
> The layered deployment may miss these information, so couldn't be used to build your solutions. However, you can always commit this layered deployment to your repo along with the traditional deployment manifest. This won't impact the build process.

@blackchoey Thanks for your reply! Yes I am aware of the limitation you mentioned. The issue is that we don't manage the base (full) deployment and only intend to build our custom modules and deliver a layered deployment manifest. Isn't that the whole point of having layered deployment? i.e. decouple modules from base deployment.

For the time being I wrote a script to merge the layered_deployment_template.json and base_deployment_template.json(an arbitrary version got from another team) , build and split it again to produce the final layered_deployment.json. You can see it's not neat at all. A better solution could definitely integrated into iotedgedev.

[...]

PR #494, opened 2021-08-11
Add push and build for layered deployments

EliiseS:

- [x] Unit tests pass locally

## Additional comments

Part of #430, but not the full solution. (Don't close the issue with this PR)

This PR covers:

- Adding ability to create config files from layered deployment templates.
- Push and build images from layered deployment templates.
- Minor fixes to the dev environment (linting, small refactoring)
- Pin all python versions to prevent https://github.com/pypa/pip/issues/9215
- Fix issues with tests generating non git ignored code and not cleaning up after themselves

> Layered deployment template reference: https://github.com/MicrosoftDocs/azure-docs/blob/master/articles/iot-edge/how-to-deploy-cli-at-scale.md#configure-a-deployment-manifest

Layered deployment with no modules: https://github.com/Azure/iotedge/issues/5299


setup.py
```diff
@@ -38,7 +32,8 @@
 setup(
     name='iotedgedev',
     version='3.1.2',
```
EliiseS, line 34: Should I pump the version?

.devcontainer/devcontainer.json
```diff
@@ -10,6 +10,7 @@
         "--network",
         "host"
     ],
+    "postCreateCommand": "make install && pip install -e .",
```
marianan, line 13: >make install && pip install -e . [](http://example.com/codeflow?start=26&length=32)

I don't know the difference between these 2 commands but I guess we need only one at the time?

.env.tmp
```diff
@@ -31,6 +31,7 @@ CONTAINER_REGISTRY_PASSWORD=""
 #
 
 EDGE_RUNTIME_VERSION="1.0"
+LAYERED_RUNTIME_TAG="1.0"
```
marianan, line 34: >LAYERED_RUNTIME_TAG [](http://example.com/codeflow?start=0&length=19)

Should this be a boolean flag instead of integer version? Also, I would assume the default value should be false?

marianan (changes requested):
:clock1:

tests/test_iotedgedev_layereddeployments.py
```diff
@@ -0,0 +1,24 @@
...
+
+pytestmark = pytest.mark.unit
+
+tests_dir = os.path.join(os.getcwd(), "tests")
+test_assets_dir = os.path.join(tests_dir, "assets")
+test_file_1 = os.path.join(test_assets_dir, "layered_deployment.template.json")
+test_solution_shared_lib_dir = os.path.join(tests_dir, "assets", "test_solution_shared_lib")
+
+
+def test_build_and_push():
```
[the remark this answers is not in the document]
EliiseS: The config wouldn't look different for the build and push of these. Base deployment config is a regular deployment config and layered deployments all look like these. There will be more interesting test cases once I add deployment functionality too.

.env.tmp
EliiseS: My plan was to use this for the layered deployment versions, since they need to be used versions. On that note, I should rename this though, maybe LAYERED_DEPLOYMENT_VERSION?
We could add ENABLE_LAYERED deployment bool in addition, but I wasn't sure since it feels a bit clunky.

iotedgedev/deploymentmanifest.py
```diff
@@ -151,25 +157,46 @@ def validate_deployment_template(self):
...
+            if self.is_layered_deployment_schema():
+                validation_success = self._validate_deployment_manifest_schema()
             validation_success &= self._validate_create_options()
         except Exception as err:
             self.output.info("Unexpected error during deployment manifest validation, skip the validation. Error:%s" % err)
 
         return validation_success
 
+    def is_layered_deployment_schema(self):
+        return "content" in self.json
```
marianan, line 169: >"content" [](http://example.com/codeflow?start=15&length=9)

I wonder if there is a schema definition we could leverage here, similar to [https://github.com/marianan/schemastore/blob/master/src/schemas/json/azure-iot-edge-deployment-template-4.0.json](https://github.com/marianan/schemastore/blob/master/src/schemas/json/azure-iot-edge-deployment-template-4.0.json)

iotedgedev/deploymentmanifest.py
```diff
@@ -151,25 +157,46 @@ def validate_deployment_template(self):
...
 
+    def _get_module_content_split(self):
+        return DeploymentManifest.dot_to_json(self._get_module_content())
+
     def _get_module_content(self):
         if "modulesContent" in self.json:
             return self.json["modulesContent"]
         elif "moduleContent" in self.json:
             return self.json["moduleContent"]
+        elif "content" in self.json and "modulesContent" in self.json["content"]:
```
marianan, line 183: >"content" in self.json [](http://example.com/codeflow?start=13&length=22)

Do you think that if the code uses is_layered_deployment_schema() instead here it will give more context to the reader?

tests/test_iotedgedev_layereddeployments.py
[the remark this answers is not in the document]
marianan: You are right, i don't think build scenarios have much of a difference between layered and regular deployments except the root content element. Adding @veyalla - Venkat - could you take a look at the test cases that Eliise added for layered deployment build/push scenarios and advise if we should add any other cases?

marianan (approved):
:shipit:

setup.py
marianan: Yes, let's update the version. I wonder if we should work in a feature branch until we are ready with all changes we want to publish. What do you think? Or should we publish each use case under a separate version?

iotedgedev/deploymentmanifest.py
EliiseS: I'm not sure that would work, since the schemas I have start with `content` and this requires `modulesContent` as the first prop (If  I'm reading it right

iotedgedev/deploymentmanifest.py
EliiseS: Could change it to:
```suggestion
        elif self.is_layered_deployment_schema() and "modulesContent" in self.json["content"]:
```

setup.py
marianan: Let's update the version when we are ready to release. For now, keep it as it is.

PR #502, opened 2021-09-01
Update documentation with iothub deploy command

EliiseS:

- [ ] Unit tests pass locally

## Description

As discussed, adding documentation for how to use the new iothub deploy command. Let me know if there's anything I've missed.

### DIFF
--- a/docs/Usage.md
+++ b/docs/Usage.md
@@ -10,7 +10,9 @@ Note: See the ["Test Coverage"](test-coverage) wiki page to see what the IoT Edg
 
 1. **Initialize IoT Edge solution and setup Azure resources**
 
-    `iotedgedev init`
+    ```sh
+    iotedgedev init
+    ```
 
     `iotedgedev init` will run both `iotedgedev new .` and `iotedgedev iothub setup`, which will create a new solution and setup your Azure resources in a single command.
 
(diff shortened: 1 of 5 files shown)


docs/Usage.md
```diff
@@ -56,28 +58,32 @@ If you are using a separate Edge device, like a Raspberry Pi, you do not need to
...
+    * You are on Windows or macOS.
+    * You are on Linux, and you have followed the [post-installation steps for Linux](https://docs.docker.com/install/linux/linux-postinstall/#manage-docker-as-a-non-root-user) to allow running Docker commands without `sudo`.
 
 1. **Deploy modules to IoT Edge device**
 
     `iotedgedev deploy`
-    > You can also combine push and deploy with `iotedgedev push --deploy`
+    You can also combine push and deploy with `iotedgedev push --deploy`
+
+1. **Create deployments on IoT Hub**
```
marianan, line 83: as iotedge uses layered deployment as a concept rather than iothub deployments, do you think we should add that term int he documentation explicitely?

iotedgedev/cli.py
```diff
@@ -262,19 +262,24 @@ def deploy(manifest_file):
 @click.option("--name",
               "-n",
               required=True,
-              help="Specify the deployment name")
+              help=("Specify the name of the deployment that will be created in the IoT hub. "
+                    "Give your deployment a unique name that is up to 128 lowercase letters. Avoid spaces and the following invalid characters: & ^ [ ] { } \\ | \" < > /."))
```
marianan, line 266: >that is up to 128 lowercase letters. Avoid spaces and the following invalid characters: & ^ [ ] { } \\ | \" < > /." [](http://example.com/codeflow?start=56&length=115)

should we instead point to the msdn doc that lists these limitations in case they change in the future? iotedgedev isn't doing the validation explciitely so I thought we should delegate the definition too :)

EliiseS: Right, good point, I'll replace with link

docs/Usage.md
EliiseS: I've editing to the reference 'automatic deployments' (what they call it)

marianan (approved):
:shipit:

PR #526, opened 2022-03-11
Vikas/azure

vikas0212git:


azure-pipelines.yml
```diff
@@ -65,16 +80,24 @@ steps:
...
-- task: Bash@3  
-  condition: and(succeeded(), startsWith(variables['Build.SourceBranch'], 'refs/tags/'))
+- task: CopyFiles@2
+  inputs:
+    SourceFolder: '$(Build.SourcesDirectory)/dist/'
+    Contents: '*'
+    TargetFolder: '$(Build.ArtifactStagingDirectory)'
+
+- task: Bash@3
+  displayName: 'Uplaod to Blob'
```
marianan, line 90: >Uplaod [](http://example.com/codeflow?start=16&length=6)

nit: typo

azure-pipelines.yml
```diff
@@ -87,10 +110,10 @@ steps:
     addChangeLog: false
 
 - task: Bash@3
-  displayName: "Verify wheel"
+  displayName: "Install wheel"
   inputs:
     targetType: 'inline'
-    script: pip install --upgrade iotedgedev && iotedgedev --version
+    script: pip install dist/iotedgedev-${VERSION}-py3-none-any.whl && iotedgedev --version
```
marianan, line 116: >script: pip install dist/iotedgedev-${VERSION}-py3-none-any.whl && iotedgedev --version [](http://example.com/codeflow?start=4&length=87)

this step should be right after l52 i think to validate the locally built wheel? Because this pipeline no longer publishes to pypi directly we cannot validate the install from pypi as it was the original intention.

marianan (changes requested):
:clock1:

azure-pipelines.yml
```diff
@@ -86,15 +88,25 @@ steps:
...
+#   displayName: 'Upload to Blob'
+#   condition: and(succeeded(), eq(variables['WHEEL_PUBLISH_FLAG'], 'true'))
+#   inputs:
+#     targetType: 'inline'
+#     script: |
+#       export AZCOPY_SPA_CLIENT_SECRET=$(AZURESDKPARTNERDROPS-SERVICE-PRINCIPAL-KEY)
+#       azcopy login --service-principal --application-id $(APPID)
+#       azcopy copy "$(Build.ArtifactStagingDirectory)/*"  "https://azuresdkpartnerdrops.blob.core.windows.net/drops/azure-iot-edge-tools-iotedgedev/pypi/$(VERSION)/"
+
 - task: Bash@3
```
mhshami01, line 101: If it's not too much trouble, I would prefer to use 'AzureFileCopy@4' - similar to IoTEdgeHubDev.

azure-pipelines.yml
```diff
@@ -86,15 +88,25 @@ steps:
...
-  displayName: 'Uplaod to Blob'
+  displayName: 'Upload to Blob'
   condition: and(succeeded(), eq(variables['WHEEL_PUBLISH_FLAG'], 'true'))
   inputs:
     targetType: 'inline'
     script: |
       export AZCOPY_SPA_CLIENT_SECRET=$(AZURESDKPARTNERDROPS-SERVICE-PRINCIPAL-KEY)
       azcopy login --service-principal --application-id $(APPID)
-      azcopy copy "$(Build.ArtifactStagingDirectory)/*"  "https://azuresdkpartnerdrops.blob.core.windows.net/drops/azure-iot-edge-tools-iotedgedev/pypi/$(VERSION)/"
+      azcopy copy "$(Build.ArtifactStagingDirectory)/*"  "https://sadevicelabsetup.blob.core.windows.net/test//azure-iot-edge-tools-iotedgedev/pypi/$(VERSION)/"
```
mhshami01, line 109: We used 'python' for IoTEdgeHubDev. Is 'pypi' the better option? Either way, I think we should be consistent.

azure-pipelines.yml
vikas0212git: I have removed the Install wheel task and added it to the build wheel.

azure-pipelines.yml
```diff
@@ -109,11 +121,11 @@ steps:
...
-  inputs:
-    targetType: 'inline'
-    script: pip install dist/iotedgedev-${VERSION}-py3-none-any.whl && iotedgedev --version
+# - task: Bash@3
+#   displayName: "Install wheel"
+#   inputs:
+#     targetType: 'inline'
+#     script: pip install dist/iotedgedev-${VERSION}-py3-none-any.whl && iotedgedev --version
 
 - task: Bash@3
```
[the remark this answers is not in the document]
vikas0212git: I have deleted the Install wheel, not sure to remove the upload to acr task

azure-pipelines.yml
vikas0212git: AzureFileCopy@4 is not supported on the ubuntu vm

azure-pipelines.yml
vikas0212git: yes, I have updated to python.

azure-pipelines.yml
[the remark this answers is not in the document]
marianan: Please do not delete the upload to acr task. We publish an mcr image (mcr.microsoft.com/iotedge/iotedgedev) which has iotedgedev preinstalled and ready to use. I don't know the usage numbers on it but this is different from the docker image with the Edge runtime that I deleted earlier.

marianan (approved):
:shipit:

mhshami01 (approved):

PR #536, opened 2022-04-22
Install azure-cli with pip

beandrad:
Installing it with apt-get was causing issues when running some commands
that need uamqp library.

Note that since the python docker image already has `setuptools== 58.1.0`,
we needed to downgrade to a version < 58.0.0 to make it compatible with
azure-cli==2.25.0 (compatible with azure-cli-core==2.25.0 pinned
in this project). See related [issue](https://github.com/Azure/azure-cli/issues/19468).

- [ ] Unit tests pass locally

## Additional comments

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -28,13 +28,6 @@ RUN \
     curl -fsSL https://deb.nodesource.com/setup_${NODE_VERSION} | bash - && \
     apt-get install -y nodejs && \
     npm install npm@latest -g && \
-    # Install azure-cli
-    apt-get install -y ca-certificates curl apt-transport-https lsb-release gnupg && \
-    curl -sL https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor | tee /etc/apt/trusted.gpg.d/microsoft.gpg > /dev/null && \
-    echo "deb [arch=amd64] https://packages.microsoft.com/repos/azure-cli/ $(lsb_release -cs) main" | tee /etc/apt/sources.list.d/azure-cli.list && \
-    apt-get update && apt-get install -y azure-cli=${AZURE_CLI_VERSION} && \
-    az extension add --name azure-iot --system && \
-    az extension update --name azure-iot && \
     # Use Docker script from script library to set things up - enable non-root docker, user vscode, using moby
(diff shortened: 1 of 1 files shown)


.devcontainer/Dockerfile
```diff
@@ -58,6 +51,10 @@ RUN \
     rm -rf /tmp/* && \
     rm -rf /var/lib/apt/lists/*
 
+# Install azure-cli
+RUN pip install setuptools==57.0.0 \
```
marianan, line 55: >setuptools==57.0.0 [](http://example.com/codeflow?start=16&length=18)

i believe that this dependency is managed internally by the az cli?

beandrad: It is true that az cli installs `setuptools==52.0.0` [requirements.txt](https://github.com/Azure/azure-cli/blob/162be08eefaa1d64974a07b6179babe4c5b2830a/requirements.txt#L2). However, we still need to make sure that the version of setuptools is <58.0.0 because of the following:
- The installation of python3.9 includes a newer version of setuptools (incompatible with az cli 2.25.0);
- az cli  has a dependency on `pygithub==1.38`, which may get installed before `setuptools` is downgraded to 52.0.0 (please note that there's no way to specify the order of installation with pip);
- this version of `pygithub==1.38` needs setuptools <58.0.0 because of a deprecated featured (see [code](https://github.com/PyGithub/PyGithub/blob/v1.38/setup.py#L104)).

Ideally, we should be using newer versions of az cli, but `azure-cli-core` is pinned to 2.25.0 (see [code](https://github.com/Azure/iotedgedev/blob/55a4883a5006e85ccb3d1f33195feae803b7dc78/requirements.txt#L10)).

marianan: I'd recommend to use a newer version of az cli instead of trying to account for deprecated lib versions.

beandrad: Yes, that would be ideal, but as mentioned, `iotedgedev` pins `azure-cli-core` to 2.25.0 (see [code](https://github.com/Azure/iotedgedev/blob/55a4883a5006e85ccb3d1f33195feae803b7dc78/requirements.txt#L10)), which means that  the version of az-cli installed via pip needs to be the same.

Issue #560, opened 2022-05-10
Update azure-cli-core


beandrad:
The azure-cli-core dependency is currently pinned to version 2.25.0, which is quite old, so it should be updated. Apart from that, if we want to install azure-cli when within the same environment using pip, the same version of azure-cli needs to be used. This version has a dependency on an old library that makes use of a deprecated feature of `setuptools`. Therefore, if we want to set the development environment this way, we need to use an old `setuptools` library.

azure-cli-core should be pinned to version 2.34.1.  Please note >=2.35.0 cannot be used as this version is not compatible with the `docker` dependency, since `docker` requires websocket-client==0.56.0 and azure-cli-core>=2.35.0 requires websocket-client==1.31.1.

shiroyama:
I'd like to join this

PR #568, opened 2022-05-17
Update azure-cli-core

shiroyama:
## Checklist

- [x] Versions update reflected in all places (both __init__.py files, CHANGELOG, setup.py, setup.cfg)
- [x] Unit tests pass locally

### General guidelines

- [x] Title of the pull request is clear and informative.
- [x] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR.

## Description

The azure-cli-core dependency is currently pinned to version 2.25.0,
which is quite old, so it should be updated. Apart from that, if we want
to install azure-cli when within the same environment using pip, the
same version of azure-cli needs to be used. This version has a
dependency on an old library that makes use of a deprecated feature of
setuptools. Therefore, if we want to set the development environment
this way, we need to use an old setuptools library.

azure-cli-core should be pinned to version 2.34.1. Please note >=2.35.0
cannot be used as this version is not compatible with the docker
dependency, since docker requires websocket-client==0.56.0 and
azure-cli-core>=2.35.0 requires websocket-client==1.31.1.

Please note we removed the manifest for no_modules and the related test
cases because the validation of iot edge deployment create got stricter
with the current version and otherwise the test failed.

Fixes: #560

Co-authored-by: Isabel Lopez Andrade <beandrad@microsoft.com>


setup.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 3.3.4
+current_version = 3.3.5
```
marianan, line 2: >.5 [](http://example.com/codeflow?start=21&length=2)

Please do not upgrade the iotedgedev version just yet. We will publish all the Hackathon changes at once at the end of the week and will bump the version then.

setup.py
```diff
@@ -12,7 +12,7 @@
     'python-dotenv',
     'requests >= 2.20.0, <= 2.25.1',
     'fstrings',
-    'azure-cli-core >= 2.25.0',
+    'azure-cli-core >= 2.34.1, < 2.35.0',
```
marianan, line 15: >'azure-cli-core >= 2.34.1, < 2.35.0', [](http://example.com/codeflow?start=4&length=37)

I copied this from the PR description, it would be great to add this info for future context for this line.
"Please note >=2.35.0
cannot be used as this version is not compatible with the docker
dependency, since docker requires websocket-client==0.56.0 and
azure-cli-core>=2.35.0 requires websocket-client==1.31.1."

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to this project since 0.82.0 will be documented in this file.
 
+## [3.3.5] - 2022-05-17
+- Update azure-cli-core
```
marianan, line 5: In general, we will publish a version update for iotedgedev with all the changes from the Hackathon, so continue appending the changes in the Changelog but don't bump the version in the code.

PR #571, opened 2022-05-18
Update azure-cli-core

### LINKED ISSUE #560: Update azure-cli-core

beandrad:
## Checklist

- [x] Unit tests pass locally

### General guidelines

- [ ] Title of the pull request is clear and informative.
- [ ] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR.

## Description

The azure-cli-core dependency is currently pinned to version 2.25.0,
which is quite old, so it should be updated. Apart from that, if we want
to install azure-cli when within the same environment using pip, the
same version of azure-cli needs to be used. This version has a
dependency on an old library that makes use of a deprecated feature of
setuptools. Therefore, if we want to set the development environment
this way, we need to use an old setuptools library.

azure-cli-core should be pinned to version 2.34.1. Please note >=2.35.0
cannot be used as this version is not compatible with the docker
dependency, since docker requires websocket-client==0.56.0 and
azure-cli-core>=2.35.0 requires websocket-client==1.31.1.

Please note we removed the manifest for no_modules and the related test
cases because the validation of iot edge deployment create got stricter
with the current version and otherwise the test failed.

Fixes #560


tests/test_iotedgedev_iothub_deploy.py
```diff
@@ -1,4 +1,5 @@
 import os
+import shutil
```
marianan, line 2: >shutil [](http://example.com/codeflow?start=7&length=6)

what scenario needs this library?

beandrad: forgot to remove, sorry!

marianan (approved):
:shipit:

tests/test_iotedgedev_solution_build.py
```diff
@@ -69,25 +69,6 @@ def test_solution_build_and_push_with_layered_deployment():
     assert 'ERROR' not in result.output
 
 
-def test_solution_build_and_push_with_layered_deployment_no_modules():
```
mhshami01, line 72: Question: Would please explain why this test is not needed?

beandrad: Please note we removed the manifest for no_modules and the related test
cases because the validation of iot edge deployment create got stricter
with the current version and otherwise the test failed.

PR #574, opened 2022-05-18
Deprecate duplicate commands

### LINKED ISSUE #559: Deprecate duplicate commands
Some commands can be called in two different ways (for instance, `iotedgedev init` and `iotedgev solution init`). Eventually, this duplication should be removed to simplify the tool. This is a breaking change, though; so before implementing it in the next major release, we should add a deprecation notice in such commands. This deprecation notice should be added to the help message and printed when the command is run. This can be done by setting `deprecated=True` in the `click.Command()` decorator (see [documentation](https://click.palletsprojects.com/en/8.1.x/api/#click.Command)).

Note that the reason for this duplication is that the commands are added to two `click` groups (`main` and `solution`). The commands that should be deprecated are the ones corresponding to the  `main` group.

shiroyama:
## Checklist

- [ ] Unit tests pass locally

### General guidelines

- [ ] Title of the pull request is clear and informative.
- [ ] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR.

## Description

Some commands can be called in two different ways (for instance,
`iotedgedev init` and `iotedgev solution init`). Eventually, this
duplication should be removed to simplify the tool. This is a breaking
change, though; so before implementing it in the next major release, we
should add a deprecation notice in such commands. This deprecation
notice should be added to the help message and printed when the command
is run. This can be done by setting `deprecated=True` in the
`click.Command()` decorator (see
[documentation](https://click.palletsprojects.com/en/8.1.x/api/#click.Command)).

Note that the reason for this duplication is that the commands are added
to two `click` groups (`main` and `solution`). The commands that should be
deprecated are the ones corresponding to the `main` group.

Fixes #559

### DIFF
--- a/iotedgedev/cli.py
+++ b/iotedgedev/cli.py
@@ -3,6 +3,11 @@ import socket
 import sys
 
 import click
+import copy
+import typing as t
+
+from click import Command
+from click import Group
 from fstrings import f
 
 from .azurecli import AzureCli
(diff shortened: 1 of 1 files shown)


marianan (approved):
:shipit:

marianan:
looked into the pipeline failures briefly, it may be that 'deprecated' library needs to be added into requirements.txt

shiroyama:
@marianan My apologies, the `deprecated` was accidentally added to the import list. By removing it everything seems working. Now the pipeline is all green. It'd be really nice if you could take a look again! Cheers!

PR #585, opened 2022-07-05
fix: duplicate prompt messages

### LINKED ISSUE #558: Remove dupe log messages in `iotedgedev init`
- iotedgedev Version: v3.3.4
- Python Version: 3.9
- Pip Version: 22.0.4
- Development machine OS Version: Debian GNU/Linux 11 (bullseye)
- IoT Edge device OS Version: Debian GNU/Linux 11 (bullseye)

The `iotedgedev init` print some messages twice:

Screenshot 2022-05-07 at 11 12 21

Steps to Reproduce:

1. Run `iotedgedev init`

roel4ez:
## Checklist

- [ ] Unit tests pass locally

### General guidelines

- [x] Title of the pull request is clear and informative.
- [x] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR.

## Description

Fixes #558 

hi @marianan - This PR includes a commit to showcase the removal of the duplicate prompt message (only for subscription list right now). Let me know if you are ok with using this approach, and then I will work out the rest of the duplicate prompts.

### Additional information

The problem seems to be that the method that sets the defaults is the one printing messages and when the `prompt` option is also set for the parameter, the default function gets called twice (see [code](https://github.com/pallets/click/blob/c96545f6f4ba0eab99de6ec8b4ceb77c9bdb2528/src/click/core.py#L2835)).

The solution used here is to prompt the user in the default method instead of using the `prompt` click option. 

Thanks @beandrad for the help


marianan:
@roel4ez thanks for the PR and apologies for the late reply. I am ok with the approach, feel free to apply tot he rest of the commands.

roel4ez:
> @roel4ez thanks for the PR and apologies for the late reply. I am ok with the approach, feel free to apply to the rest of the commands.

@marianan updated and applied to all arguments.

iotedgedev/cli.py
```diff
@@ -628,25 +628,32 @@ def validate_option(ctx, param, value):
...
+    return input() or iothub_name
 
 
 def list_resource_groups_and_set_default():
     if not azure_cli.list_resource_groups():
         sys.exit()
-    return "iotedgedev-rg"
+
+    output.info("Enter Resource Group Name (Creates a new Resource Group if not found) [iotedgedev-rg]:")
+    resource_group = input() or "iotedgedev-rg"
```
marianan, line 655: >resource_group = [](http://example.com/codeflow?start=4&length=16)

could just return here instead of creating resource_group

iotedgedev/cli.py
```diff
@@ -663,14 +670,18 @@ def list_subscriptions_and_set_default():
     if not azure_cli.list_subscriptions():
         sys.exit()
     default_subscriptionId = azure_cli.get_default_subscription()
-    return default_subscriptionId
+    output.info(f"Select an Azure Subscription Name or Id [{default_subscriptionId}]:")
+    return input() or default_subscriptionId
```
marianan, line 674: >default_subscriptionId [](http://example.com/codeflow?start=22&length=22)

azure_cli.get_default_subscription(), no need for intermediate variable

iotedgedev/cli.py
```diff
@@ -663,14 +670,18 @@ def list_subscriptions_and_set_default():
...
 
 
-def header_and_default(header, default, default2=None):
+def header_and_default(header, prompt, default, default2=None, options=None):
     output.header(header)
     if default == '' and default2 is not None:
         return default2
```
marianan, line 680: >return default2 [](http://example.com/codeflow?start=8&length=15)

if the code returns here it will miss the input prompt. Let's use an intermediate variable to capture the return value

marianan (changes requested):
:clock1:

iotedgedev/cli.py
roel4ez: not sure, since `azure_cli.get_default_subscription()` actually calls `az account show`, I think it may be better to execute that only once. wdyt?

iotedgedev/cli.py
roel4ez: I think the `default2` is confusing, and is only used at line 729. I think it may be better to remove the `default2` argument, and instead change line 729 to something like
`default=lambda: header_and_default('RESOURCE GROUP LOCATION', 'Enter a Resource Group Location', envvars.RESOURCE_GROUP_LOCATION or 'westus',` so it chooses `westus` as default if the environment variable is empty, which I assume is the goal of `default2`

wdyt?

iotedgedev/cli.py
marianan: if we want to execute az accoutn show once and preserve the azure_cli abstraction layer i think we should cache the response the first time when it is called. I am happy to leave it for a separate PR too.

iotedgedev/cli.py
marianan: your proposal to update l 729 sounds good to me

roel4ez: fixed in https://github.com/Azure/iotedgedev/pull/585/commits/3b9a769a647e2c14096c778d81e887765704bc90

iotedgedev/cli.py
marianan: yes, feel free to create a tracking issue, thanks so much

PR #636, opened 2026-08-06
Refresh CI agent images and update auth methods for tests

DrewH-ms:
## Why

The CI agent images (`windows-2019`, `ubuntu-20.04`) have been retired, so every run fails at agent allocation:

```
No image label found to route agent pool Azure Pipelines.
```

`Azure.iotedgedev` is a required check, so this blocks all merges.

## What

Refreshes the images and fixes what breaks on the newer agents. No product code changes.

- **Images** — `windows-2022`, `ubuntu-22.04`. Linux 3.6 leg dropped (not available on 22.04; Windows still covers 3.6). Linux now runs even if Windows fails.
- **Test auth** — adds optional Entra ID auth via an ARM service connection, instead of credentials in pipeline variables. Opt-in through `IOTHUB_AUTH_TYPE=login`; unset behaves exactly as today.
- **Test fixtures** — .NET 8 / `ltsc2022`. `dotnet/sdk:2.1` has no Windows image for the 2022 agent kernel, so builds fail with an OS mismatch.
- **Deps** — declare `jsonschema` (imported by `deploymentmanifest.py` but never declared; unpinned because current releases need Python ≥ 3.10). Drop unused `msrestazure` and `docker-compose`.
- **Misc** — `nanoserver` from MCR; `azure-pipelines.yml` cleanup (unwired file, happy to drop it from this PR).

## Validation

All 7 legs green against this exact tree.

Two notes:
- The service connection wiring is exercised for the first time by this PR's own run — may need a one-time resource authorization, and the `azureServiceConnection` default may need to match this project's connection name.
- Container scan and `CFS0011` warnings are pre-existing and identical on `main`.


setup.py
```diff
@@ -15,6 +15,7 @@
     # Note >=2.35.0 cannot be used as is not compatible with the docker dependency;
     # docker requires websocket-client==0.56.0 and azure-cli-core>=2.35.0 requires websocket-client==1.31.1.
     'azure-cli-core >= 2.34.1, < 2.35.0',
+    'jsonschema',
```
c-ryan-k: we may want to pin this to at least a major version range, it has bit us CLI-side in the past

requirements.txt
```diff
@@ -3,10 +3,10 @@ docker==5.0.3
 python-dotenv==0.19.0
 requests==2.25.1
 fstrings==0.1.0
-msrestazure==0.6.4
 iotedgehubdev==0.14.18
 applicationinsights==0.11.9
 commentjson==0.9.0
 azure-cli-core==2.34.1
+jsonschema
```
c-ryan-k: same comment here, can we pin a version?

vsts_ci/win32/continuous-build-win32.yml
```diff
@@ -15,12 +24,37 @@ steps:
...
+  - ${{ else }}:
+      - task: AzureCLI@2
+        displayName: "Run tests against iotedgedev source code (AAD)"
+        inputs:
+          azureSubscription: ${{ parameters.azureServiceConnection }}
+          scriptType: ps
+          scriptLocation: inlineScript
+          inlineScript: |
+            az --version
+            az extension add --name azure-iot --only-show-errors
```
c-ryan-k: do we want to specify a version of `azure-iot` here? this will currently always grab the newest version released ot the extension index, just making sure that's expected.  One more tiny nit, `extension add` can fail if the extension is already on the agent for whatever reason (likely not an issue here if this passes 😄 ) but `extension upgrade --add` will always either install or upgrade regardless.

tests/utility.py
```diff
@@ -34,7 +34,10 @@ def get_docker_client():
     envvars.load(force=True)
     utility = Utility(envvars, output)
     docker_client = Docker(envvars, utility, output)
-    docker_client.init_registry()
+    try:
+        docker_client.init_registry()
```
c-ryan-k, line 38: I'm sure this was added to suppress an exception that deserves to be, but would there be any exceptions here that we would want to throw instead of just `pass`? `try-except-pass` is usually a linter rule in most of our repos to ensure we're not swallowing exceptions we don't anticipate

c-ryan-k (approved):

PR #639, opened 2026-09-11
IoTEdgeDev Python 311-314 upgrade

DrewH-ms:
Upgrades `python_requires` from `>=3.6, <3.10` to `>=3.11, <3.15` and moves CI to a 3.11–3.14 matrix. **No product code changes** — `git diff main..HEAD -- iotedgedev/` is empty.

- **`setup.py`** — new Python range and classifiers; dependency floors raised. Removes the upper caps that made the tree unresolvable on modern Python (`azure-cli-core < 2.35.0`, `requests <= 2.25.1`, `more-itertools < 8.1.0`, `bcrypt <= 3.1.7`). Swaps `pypiwin32` → `pywin32` and `bumpversion` → `bump2version` (the old release dies on 3.12 via `ConfigParser.readfp()`).
- **`vsts_ci/`** — matrix is now 3.11–3.14 on `windows-2022`, `ubuntu-22.04`, `ubuntu-24.04`. Tests authenticate to the test registry with a short-lived, registry-scoped AAD token instead of admin credentials, which stay disabled.
- **`azure-pipelines.yml`** — release build targets 3.11.
- **Docs / devcontainer** — `CONTRIBUTING.md`, setup docs and `.devcontainer` updated to the supported range.

Version bumped to **3.3.9** (patch — no product behavior change). `3.3.8` was set on `main` but never published to PyPI, where the latest release remains `3.3.7`.

**Release Docker image**

`docker/tool/linux` no longer built at all on `main` — the NodeSource `node_14.x` signing key was retired and the Maven 3.5.4 URL now 404s — and its Python 3.9 could not have installed the new wheel. Re-based on Ubuntu 24.04, whose default `python3` is 3.12:

- Node 14 → 20, .NET SDK 2.1 → 8.0, Maven 3.5.4 → 3.8.7 (apt, no tarball), Java 8 retained.
- `docker-ce` → `docker-ce-cli` + `docker-compose-plugin`; the container talks to the host daemon over the mounted socket, so the daemon package was dead weight.
- `iotedgedev` installs into a virtualenv at `/opt/iotedgedev`. Ubuntu 24.04 is PEP 668 managed, so the previous `sudo pip3 install` of the wheel would fail outright.

The step that was supposed to catch this ran `docker run` against an image with no `CMD`: the container started, exited 0, and passed unconditionally. It now asserts the interpreter is in range, prints every tool version, and scaffolds C#, Node.js, Python and Java modules. `docs/test-coverage.md` was also years out of date and now matches the real matrix.

**Verification**

- CI build **8099**: **12/12 jobs green**, `119 passed, 1 skipped` per job.
- Release build **8102**: builds and installs the wheel end to end under the new `Requires-Python`.
- Locally on **Python 3.14**: `pip install -e .` succeeds and `pytest -m unit` gives **73 passed**.
- Docker image: toolchain layers build clean and report python 3.12.3, dotnet 8.0.131, node 20.20.2, mvn 3.8.7, java 1.8.0_502. The wheel installs into the venv; against the system interpreter the same install fails with `externally-managed-environment`, which is the behaviour the venv exists to avoid. The npm and PyPI layers can only be exercised on a build agent — corporate egress blocks both from a local container.


docker/tool/linux/Dockerfile.base
```diff
@@ -1,64 +1,74 @@
...
-ENV DEBIAN_FRONTEND noninteractive
-ENV DOTNETCORESDK_VERSION 2.1
-ENV DOCKER_COMPOSE_VERSION 1.22.0
-ENV MAVEN_VERSION=3.5.4
-RUN apt-get update && apt-get install -y apt-transport-https ca-certificates curl software-properties-common && \
-    curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add - && \
-    add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" && \
+FROM mcr.microsoft.com/mirror/docker/library/ubuntu:24.04
+ENV DEBIAN_FRONTEND=noninteractive
+ENV DOTNETCORESDK_VERSION=8.0
```
damonbarry: .NET 8 goes out of support in November. How much effort would it be to snap to .NET 10 instead?

DrewH-ms: Good catch, not much so I bumped it

damonbarry (approved):

